PR #17, opened 2018-03-22
Create the Upload letters task

banderous:
### Change description ###

Created an UploadLetters task to upload letters in the Created state (note that the task is not currently scheduled and will not be executed).

This required changing the Letter entity to use getters and setters instead of public final fields, to allow fields to be updated.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLetters.java
```diff
@@ -0,0 +1,55 @@
...
+import uk.gov.hmcts.reform.sendletter.entity.LetterState;
+import uk.gov.hmcts.reform.slc.services.steps.getpdf.FileNameHelper;
+import uk.gov.hmcts.reform.slc.services.steps.getpdf.PdfDoc;
+import uk.gov.hmcts.reform.slc.services.steps.sftpupload.FtpClient;
+import uk.gov.hmcts.reform.slc.services.steps.sftpupload.exceptions.FtpStepException;
+
+import java.sql.Timestamp;
+import java.time.Instant;
+
+public class UploadLetters {
```
rkondratowicz: May be worth adding `Task` suffix so that it's easier to find in the futures.
It will also be more consistent with current naming convention (`LettersController` etc)

src/test/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTest.java
```diff
@@ -0,0 +1,58 @@
...
+public class UploadLettersTest {
+
+    @Autowired
+    LetterRepository repository;
+
+    @Autowired
+    private EntityManager entityManager;
+
+    @Test
+    public void uploads_pdfs_to_sftp() throws Exception {
```
rkondratowicz: either update test name, or (probably better) create a separate one for checking the db update part

src/test/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTest.java
```diff
@@ -0,0 +1,58 @@
...
+        LetterService s = new LetterService(repository);
+        UUID id = s.send(SampleData.letter(), "service");
+
+        // Invoke the upload job.
+        try (LocalSftpServer f = LocalSftpServer.create()) {
```
rkondratowicz: `f`?

src/main/java/uk/gov/hmcts/reform/sendletter/entity/Letter.java
```diff
@@ -69,12 +61,58 @@ public Letter(
...
+        return state;
+    }
+
+    public void setState(LetterState state) {
+        this.state = state;
+    }
+
+    public byte[] getPdf() {
+        return pdf;
+    }
```
rkondratowicz: missing setter?

banderous: I didn't put in a setter since we don't currently need to update PDFs. The pdf is a required parameter when you create a Letter. If we need to update the PDF in future we could add a setter then.

PR #18, opened 2018-03-22
Zip pdf before sending to sftp

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/slc/services/steps/zip/ZipFileNameHelper.java
```diff
@@ -0,0 +1,24 @@
...
+
+import uk.gov.hmcts.reform.sendletter.entity.Letter;
+
+import static java.time.LocalDateTime.now;
+import static java.time.format.DateTimeFormatter.ofPattern;
+
+public class ZipFileNameHelper {
+
+    public static String generateName(Letter letter) {
+
```
banderous: whitespace

PR #21, opened 2018-03-23
Add task to mark letters as Posted

banderous:
### Change description ###

This change adds a new 'MarkLettersPosted' task that is based on the UpdateLetterStatus job from the send-letter-consumer.

An integration test is included that creates a letter in the Uploaded state, creates a dummy report on a local SFTP server, runs the task and ensures the letter is marked as Posted.

More code is copied across from the send-letter-consumer codebase including the ReportParser and its tests.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
```diff
@@ -0,0 +1,69 @@
...
+public class MarkLettersPostedTask {
+    private final LetterRepository repo;
+    private final FtpClient ftpClient;
+    private static final Logger logger = LoggerFactory.getLogger(MarkLettersPostedTask.class);
+    private final FtpAvailabilityChecker ftpAvailabilityChecker;
+    private final ReportParser parser;
+
+    @Autowired
+    public MarkLettersPostedTask(LetterRepository repo, FtpClient ftp,
+                                 FtpAvailabilityChecker checker, ReportParser parser) {
```
doncem: all arguments must be declared in same line or aligned vertically individually

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
```diff
@@ -0,0 +1,69 @@
...
+        Letter letter = repo.findById(id).orElseThrow(() -> new LetterNotFoundException(id));
+        if (LetterState.Uploaded == letter.getState()) {
+            letter.setSentToPrintAt(Timestamp.from(letterPrintStatus.printedAt.toInstant()));
+            letter.setState(LetterState.Posted);
+            repo.saveAndFlush(letter);
+            logger.info("Marking letter {} as Posted", letter.getId());
+        } else {
+            logger.info("Skipping processing of letter {} in state {}", letter.getId(), letter.getState());
+        }
+    }
```
doncem: how efficient (db call-wise) is to make batched calls and flush the repo afters (or in batches too)?

src/main/java/uk/gov/hmcts/reform/slc/services/ReportParser.java
```diff
@@ -0,0 +1,73 @@
...
+public class ReportParser {
+
+    private static final Logger logger = LoggerFactory.getLogger(ReportParser.class);
+
+    public ParsedReport parse(Report report) {
+        try (CSVParser parser = parserFor(report.content)) {
+
+            List<LetterPrintStatus> statuses =
+                stream(parser.spliterator(), false)
+                    .map(row -> toPrintStatus(row))
```
doncem: can also use reference call

src/main/java/uk/gov/hmcts/reform/slc/services/ReportParser.java
```diff
@@ -0,0 +1,73 @@
...
+                    .filter(Objects::nonNull)
+                    .collect(toList());
+
+            return new ParsedReport(report.path, statuses);
```
doncem: i think, in theory, since we only use parsed report internally, is it more efficient to use `statuses` as stream so when it gets parsed in the task runner it'll just use same parallel/sequential stream to foreach the elements.

does that make sense? will it just make things more confusing? can be reviewed at the later stage

src/main/java/uk/gov/hmcts/reform/slc/services/ReportParser.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /**
+     * Converts cvs row into a letter print status object.
+     * Returns null if conversion fails.
+     */
+    private LetterPrintStatus toPrintStatus(CSVRecord record) {
+        try {
+            return new LetterPrintStatus(
+                FileNameHelper.extractId(record.get("Filename")),
```
doncem: can this extract `UUID` instead? i think i saw `UUID.fromString` earlier in the PR. This can be caught early and returned as null from this try/catch block rather than failing to update the state

src/main/java/uk/gov/hmcts/reform/slc/services/ReportParser.java
```diff
@@ -0,0 +1,73 @@
...
+                ZonedDateTime.parse(record.get("Date") + "T" + record.get("Time") + "Z")
+            );
+        } catch (Exception exc) {
+            logger.error("Error parsing row: " + record, exc);
+            return null;
+        }
+    }
+
+    public static class ReportParsingException extends RuntimeException {
```
doncem: How precious are you with inner exception class? I would suggest to keep it separate namespace/file giving the component class such as this

src/test/java/uk/gov/hmcts/reform/sendletter/LocalSftpServer.java
```diff
@@ -19,26 +19,33 @@
 
     public static final int port = 46043;
 
-    // This is the working directory of the SFTP server.
-    public static final String pdfFolderName = "moj";
+    // These determine where our pdfs and xerox reports are stored.
+    public static final String pdfFolderName = "pdfs";
+    public static final String reportFolderName = "reports";
```
doncem: Static finals should be capital cased

src/test/java/uk/gov/hmcts/reform/sendletter/SampleData.java
```diff
@@ -27,6 +28,20 @@ public static Letter letter() throws IOException {
...
+            return new uk.gov.hmcts.reform.sendletter.entity.Letter(
+                "messageId",
+                service,
+                new ObjectMapper().readTree("{}"),
+                "a type",
+                new byte[1]
+            );
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
```
doncem: is this just to make errors unchecked?

src/main/java/uk/gov/hmcts/reform/slc/services/ReportParser.java
banderous: I agree but I copied this class over from the consumer service; are you happy for me to change it in a follow up?

src/test/java/uk/gov/hmcts/reform/sendletter/SampleData.java
banderous: Yes, I didn't want to have a checked IOException on this method since there's nothing tests could do about it than fail. Do you think it makes sense to do this?

src/main/java/uk/gov/hmcts/reform/slc/services/ReportParser.java
doncem: oh, definitely not this one. observation on the whole gave a thought

src/test/java/uk/gov/hmcts/reform/sendletter/SampleData.java
doncem: less confusing in tests, more space to name the test case. all good

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
```diff
@@ -0,0 +1,71 @@
...
+        } else {
+            logger.trace("FTP server not available, job cancelled");
+        }
+    }
+
+    private void updatePrintedAt(LetterPrintStatus letterPrintStatus) {
+        UUID id = UUID.fromString(letterPrintStatus.id);
+        Letter letter = repo.findById(id).orElseThrow(() -> new LetterNotFoundException(id));
+        if (LetterState.Uploaded == letter.getState()) {
+            letter.setSentToPrintAt(Timestamp.from(letterPrintStatus.printedAt.toInstant()));
```
rkondratowicz: `sentToPrint` is when file was uploaded to SFTP

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
```diff
@@ -0,0 +1,71 @@
...
+                    ftpClient.deleteReport(parsedReport.path);
+                });
+        } else {
+            logger.trace("FTP server not available, job cancelled");
+        }
+    }
+
+    private void updatePrintedAt(LetterPrintStatus letterPrintStatus) {
+        UUID id = UUID.fromString(letterPrintStatus.id);
+        Letter letter = repo.findById(id).orElseThrow(() -> new LetterNotFoundException(id));
```
rkondratowicz: looks like this will stop the whole process if one if the letters doesn't exist

rkondratowicz: see https://github.com/hmcts/send-letter-consumer-service/blob/master/src/test/java/uk/gov/hmcts/reform/slc/services/UpdateLetterStatusJobTest.java#L155

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
banderous: Good catch!

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
banderous: Still working on this

src/main/java/uk/gov/hmcts/reform/slc/services/ReportParser.java
banderous: This is now done, LetterPrintStatus uses UUID.

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
banderous: I've changed the call to save instead of saveAndFlush

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/entity/LetterTest.java
```diff
@@ -85,7 +70,7 @@ public void compatible_with_existing_records() throws Exception {
     @Test
     public void finds_letters_by_id_and_service() {
         repository.save(getTestLetter());
-        Letter second = getTestLetter("different");
+        Letter second = SampleData.letterEntity("different");
```
rkondratowicz: is `different` a service name or a letter type?

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/entity/LetterTest.java
```diff
@@ -32,22 +31,8 @@
...
-            );
-        } catch (IOException e) {
-            throw new RuntimeException(e);
-        }
-    }
-
     public static Letter getTestLetter() {
-        return getTestLetter("a_service");
+        return SampleData.letterEntity("a_service");
     }
```
rkondratowicz: this method can be removed

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
```diff
@@ -0,0 +1,74 @@
...
+        } else {
+            logger.trace("FTP server not available, job cancelled");
+        }
+    }
+
+    private void updatePrintedAt(LetterPrintStatus letterPrintStatus) {
+        Optional<Letter> optional = repo.findById(letterPrintStatus.id);
+        if (optional.isPresent()) {
+            Letter letter = optional.get();
+            if (LetterState.Uploaded == letter.getState()) {
```
rkondratowicz: 'yoda' condition?

src/main/java/uk/gov/hmcts/reform/slc/services/steps/getpdf/FileNameHelper.java
```diff
@@ -11,12 +13,12 @@ public static String generateName(Letter letter, String extension) {
...
     }
 
-    public static String extractId(String fileName) {
+    public static UUID extractId(String fileName) {
         String[] parts = FilenameUtils.removeExtension(fileName).split(SEPARATOR);
         if (parts.length != 3) {
             throw new UnableToExtractIdFromFileNameException();
         } else {
-            return parts[2];
+            return UUID.fromString(parts[2]);
```
rkondratowicz: what if it's `"foo"`? This should probably throw `UnableToExtractIdFromFileNameException` exception as well

rkondratowicz: also, a test would be nice.

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
banderous: As in not doing letter.getState() == LetterState.Uploaded?

Happy to change it, hangover from my c++ days.

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
```diff
@@ -0,0 +1,74 @@
...
+import static java.time.LocalTime.now;
+
+/**
+ * Fetches reports from Xerox SFTP concerning posted
+ * letters and sets posted letters as Posted in the database.
+ */
+public class MarkLettersPostedTask {
+    private final LetterRepository repo;
+    private final FtpClient ftpClient;
+    private static final Logger logger = LoggerFactory.getLogger(MarkLettersPostedTask.class);
```
rkondratowicz: a bit hidden here among actual dependencies

src/test/java/uk/gov/hmcts/reform/sendletter/services/FileNameHelperTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Test
+    public void should_throw_custom_exception_when_id_cannot_be_extracted_from_file_name() {
+        assertThatThrownBy(
+            () -> FileNameHelper.extractId("a_b.pdf")
+        ).isInstanceOf(FileNameHelper.UnableToExtractIdFromFileNameException.class);
+    }
+
+    @Test
+    public void throws_custom_exception_when_invalid_uuid() {
```
rkondratowicz: let's stick to `should...` convention in this file

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
```diff
@@ -0,0 +1,75 @@
...
+                letter.setPrintedAt(Timestamp.from(letterPrintStatus.printedAt.toInstant()));
+                letter.setState(LetterState.Posted);
+                repo.save(letter);
+                logger.info("Marking letter {} as Posted", letter.getId());
+            } else {
+                logger.info("Skipping processing of letter {} in state {}", letter.getId(), letter.getState());
+            }
+        } else {
+            logger.error("Unknown letter {}", letterPrintStatus.id);
+        }
```
doncem: does it sound productive enough if we convert this block to `Optional` specific functionality? what i mean is something as follows:

```java
optional.map(letter -> {
    if (is uploaded) {
        // do void stuff
    } else {
        // do other void stuff
    }
}).orElse(unknown letter log)
```

having second level if/else block is not entirely needed

src/main/java/uk/gov/hmcts/reform/slc/services/steps/getpdf/FileNameHelper.java
```diff
@@ -25,5 +32,12 @@ private FileNameHelper() {
...
     public static class UnableToExtractIdFromFileNameException extends RuntimeException {
 
+        public UnableToExtractIdFromFileNameException(Exception inner) {
+            super(inner);
+        }
+
+        public UnableToExtractIdFromFileNameException(String message) {
+            super(message);
+        }
     }
```
doncem: same question: how precious are you with inner exception classes?

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
banderous: We are not returning a result and would be ignoring the result of the map which would look strange, no?

Also both the map and logging lambdas would need to return values.

doncem: sigh

PR #23, opened 2018-03-23
Feature: Integration tests

doncem:
Separate unit tests from integration tests. Or vice versa on preference.

```
[ ] Yes
[x] No
```


src/integrationTest/resources/application.properties
```diff
@@ -0,0 +1,7 @@
+idam.s2s-auth.url=false
+
+servicebus.queue.connectionString=servicebusurl
```
nitinprabhu: Do we need queue properties?

src/functionalTest/resources/application.properties
```diff
@@ -1,7 +1 @@
-idam.s2s-auth.url=false
-
-servicebus.queue.connectionString=servicebusurl
-servicebus.queue.name=queuename
-
-spring.datasource.driver-class-name=org.testcontainers.jdbc.ContainerDatabaseDriver
-spring.datasource.url=jdbc:tc:postgresql:9.6://localhost/letter_tracking
+# TODO: set the configuration properties for your functional tests here
```
nitinprabhu: we could delete the file if not needed and add it when required?

src/integrationTest/resources/application.properties
doncem: ah, leftovers from before!

PR #26, opened 2018-03-26
Add SerialTaskRunner to prevent concurrently running tasks

banderous:
### Change description ###

Our background tasks should not be run concurrently on any Application servers, to avoid problems such as two tasks attempting to modify the same SFTP file.

This PR introduces the SerialTaskRunner (does that name make sense?) that uses database locks to stop the same task being run more than once at a time.

[Postgres' advisory locks](https://www.postgresql.org/docs/9.4/static/explicit-locking.html#ADVISORY-LOCKS) are used, which are designed for application use, don't involve writing to tables and are automatically released if the database connection is lost.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
```diff
@@ -0,0 +1,85 @@
...
+        this.source = source;
+    }
+
+    /**
+     * Run the task with specified ID if no other task with matching
+     * id is already running.
+     * No error is thrown if such a task is already running, the
+     * supplied Runnable is simply not executed.
+     */
+    public void tryRun(long id, Runnable runnable) throws SQLException {
```
lgonczar: How would this be invoked? Will tasks/jobs have fixed ids assigned to them?

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunnerTest.java
```diff
@@ -0,0 +1,65 @@
...
+        SerialTaskRunner.get(source).tryRun(1, runnable);
+        SerialTaskRunner.get(source).tryRun(1, runnable);
+
+        verify(runnable, times(2)).run();
+    }
+
+    @Test
+    public void does_not_run_same_task_concurrently() throws SQLException {
+        Runnable shouldNotRun = mock(Runnable.class);
+        SerialTaskRunner.get(source).tryRun(1, () -> {
```
lgonczar: nice 👍

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunnerTest.java
```diff
@@ -0,0 +1,65 @@
...
+import static org.mockito.Mockito.verify;
+
+@RunWith(SpringRunner.class)
+@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
+@DataJpaTest
+public class SerialTaskRunnerTest {
+    @Autowired DataSource source;
+
+    @Test
+    public void runs_task_if_not_running() throws SQLException {
```
rkondratowicz: ?

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
```diff
@@ -0,0 +1,85 @@
...
+        log.debug("Trying to lock {}", id);
+        try (Connection connection = source.getConnection()) {
+            boolean locked = false;
+            try {
+                if (tryLock(id, connection)) {
+                    log.debug("Acquired lock {}", id);
+                    locked = true;
+                    runnable.run();
+                } else {
+                    log.debug("Failed to acquired lock {}", id);
```
lgonczar: It would be best if we could find this entry in production logs, e.g. in case we lock ourselves forever. It would also be good to have some sort of enum, so that the Id would have a corresponding task name.

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Try to acquire a session level advisory lock with specified id without blocking.
+     *
+     * <p>A session level lock will be held until either explicitly released
+     * or the database connection is closed/dies.
+     *
+     * <p>https://www.postgresql.org/docs/9.4/static/explicit-locking.html#ADVISORY-LOCKS
+     */
+    private boolean tryLock(long id, Connection connection) throws SQLException {
```
rkondratowicz: update javadoc on return value

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
```diff
@@ -0,0 +1,85 @@
...
+        String sql = String.format("SELECT pg_try_advisory_lock(%d);", id);
+        return executeReturningBool(connection, sql);
+    }
+
+    private void unlock(long id, Connection connection) throws SQLException {
+        String sql = String.format("SELECT pg_advisory_unlock(%d);", id);
+        executeReturningBool(connection, sql);
```
rkondratowicz: this method returns void

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
banderous: Yes I was thinking having something like a 'TaskLock' enum, where each of our tasks have a value.

Another approach would be to just have a single scheduled task that wraps all of our subtasks, then none of our tasks would run concurrently (but we might not want all our tasks running on the same schedule).

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunnerTest.java
banderous: Perhaps a better name would be 'runs_task'?

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
banderous: Definitely, would log.info show up in our logs?

lgonczar: Yes, it would

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunnerTest.java
lgonczar: Or `runs_task_if_that_task_is_not_running_yet` or sth?

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
banderous: The task ids will be in an enum so we will be able to tie them back to task names.

rkondratowicz: would be much more convenient for everyone to log task name imho.
Perhaps on a next PR, when this helper is actually used (and tasks/jobs defined)

banderous: Yes, I agree

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
```diff
@@ -0,0 +1,86 @@
...
+     * @return true if lock is acquired, false otherwise
+     */
+    private boolean tryLock(long id, Connection connection) throws SQLException {
+        String sql = String.format("SELECT pg_try_advisory_lock(%d);", id);
+        return executeReturningBool(connection, sql);
+    }
+
+    private boolean unlock(long id, Connection connection) throws SQLException {
+        String sql = String.format("SELECT pg_advisory_unlock(%d);", id);
+        return executeReturningBool(connection, sql);
```
lgonczar: Is there any way of using the result of unlocking? What would it mean if it returned false? Should we log it or maybe it doesn't matter?

banderous: Good idea, I'll log it

PR #29, opened 2018-03-26
Avoid EmptyResultDataAccessException

doncem:
Making sure exception is not thrown. [documentation](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.nullability.annotations)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
@@ -50,7 +50,7 @@ public class UploadLettersTaskTest {
             // Ensure the letter is marked as uploaded in the database.
             // Clear the JPA cache to force a read.
             entityManager.clear();
-            Letter l = repository.findById(id).get();
+            Letter l = repository.findOptionalById(id).get();
             assertThat(l.getState()).isEqualTo(LetterState.Uploaded);
             assertThat(l.getSentToPrintAt()).isNotNull();
             assertThat(l.getPrintedAt()).isNull();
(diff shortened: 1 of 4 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
```diff
@@ -50,7 +50,7 @@ public void uploads_file_to_sftp_and_sets_letter_status_to_uploaded() throws Exc
             // Ensure the letter is marked as uploaded in the database.
             // Clear the JPA cache to force a read.
             entityManager.clear();
-            Letter l = repository.findById(id).get();
+            Letter l = repository.findOptionalById(id).get();
```
lgonczar: Will this compile?

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
```diff
@@ -14,7 +14,8 @@
     @Lock(LockModeType.PESSIMISTIC_WRITE)
     Stream<Letter> findByState(LetterState state);
 
-    Optional<Letter> findById(UUID id);
 
-    Optional<Letter> findByIdAndService(UUID id, String service);
+    Optional<Letter> findOptionalById(UUID id);
```
lgonczar: Is this method used? There are no code updates related with it (other than tests).

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
doncem: integration tests pass - application loads context successfully. the tests show it works

link to documentation says it is a way to avoid unnecessary exception. there would still be an error in case `id` is passed as null

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
doncem: i saw it coming in another pr

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
lgonczar: Oh, obviously. Missed the `get` at the end. And the fact that it was Optional before, as well. All fine.

PR #30, opened 2018-03-26
Handle smoke test letters

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -65,6 +68,10 @@ private void upload(Letter letter) throws IOException {
             zippedDoc.filename
         );
 
-        ftp.upload(zippedDoc);
+        ftp.upload(zippedDoc, isSmokeTest(letter));
+    }
+
+    private boolean isSmokeTest(Letter letter) {
+        return Objects.equals(letter.getType(), SMOKE_TEST_LETTER_TYPE);
```
lgonczar: I like `SMOKE_TEST_LETTER_TYPE.equals(letter.getType())` more, but it's really a matter or preference.

src/test/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Test
+    public void should_handle_smoke_test_letters() throws Exception {
+        given(zipper.zip(any(), any())).willReturn(new ZippedDoc("hello.zip", "hello".getBytes()));
+
+        givenDbContains(letterOfType(SMOKE_TEST_LETTER_TYPE));
+        task.run();
+        verify(ftpClient).upload(any(), eq(true));
+
+        givenDbContains(letterOfType("not_" + SMOKE_TEST_LETTER_TYPE));
```
lgonczar: Probably a separate test case, but I understand it was already like this in the consumer

PR #32, opened 2018-03-26
Migration script fix

lgonczar:
### Change description ###

**This pull request changes existing migration scripts, but as they haven't been applied, it's safe to make these changes**

Make the new `state` column nullable to enable the migration script to be applied on a non-empty table.

Also, change the names of newly created columns:
- `state` -> `status`, as it's the typical name
- `pdf` -> `document` to be less specific about the content

The name changes are not so important, so please let me know if you're against them.

```
[ ] Yes
[x] No
```


src/main/resources/db/migration/V006__Add_status_column.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE letters
+ADD COLUMN status VARCHAR(256);
```
rkondratowicz: fill the values here base on dates set

lgonczar:
Fair enough - I'll leave pdf as it is and make sure there's no inconsistency between db column and code for status.

src/main/resources/db/migration/V006__Add_status_column.sql
lgonczar: I'd wait with this filling this column until the service populates the status itself. Then, when we apply that script, be can be sure all records have it set properly. Otherwise, the service will create new ones without status.

rkondratowicz: it already does

lgonczar: But not the producer. The two services need to have identical sets of migration scripts.

rkondratowicz: ok

lgonczar:
Updated as requested

src/test/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
```diff
@@ -17,15 +17,18 @@
...
 
-    @Mock private LetterRepository repo;
-    @Mock private Zipper zipper;
-    @Mock private FtpClient ftpClient;
+    @Mock
+    private LetterRepository repo;
+    @Mock
+    private Zipper zipper;
+    @Mock
+    private FtpClient ftpClient;
```
rkondratowicz: :(

src/main/java/uk/gov/hmcts/reform/sendletter/data/LetterRepository.java
```diff
@@ -65,9 +64,12 @@ public void save(DbLetter letter, Instant creationTime, String messageId) throws
      * @param serviceName String
      * @return Letter status.
      */
-    public Optional<LetterStatus> getLetterStatus(UUID id, String serviceName) {
+    public Optional<uk.gov.hmcts.reform.sendletter.model.out.LetterStatus> getLetterStatus(
```
rkondratowicz: was this change intentional? now repo layer is returning api layer model

lgonczar: Nothing changed, but there are two `LetterStatus` classes used here now and one of them has be referred to its fully qualified name

src/main/resources/db/migration/V008__Populate_status_column.sql
```diff
@@ -0,0 +1,13 @@
+UPDATE
+  letters
+SET
+  status = (
+    CASE
+      WHEN sent_to_print_at IS NULL THEN
```
banderous: As discussed in the migration conversation yesterday we could set the status to 'Uploaded' (since they're on the service bus) instead of 'Created' and collapse these two cases.

This ensures the new scheduled tasks will not attempt to reupload letters in the created state that have no pdf. If the letters don't upload then our failed letters task will alert us to them.

lgonczar: Done

src/main/resources/db/migration/V008__Populate_status_column.sql
```diff
@@ -0,0 +1,11 @@
...
+  letters
+SET
+  status = (
+    CASE
+      WHEN printed_at IS NULL THEN
+        'Uploaded'
+      ELSE
+        'Posted'
+    END
+  );
```
rkondratowicz: can we put it in the same migration script where we add the column

src/main/resources/db/migration/V006__Add_status_column.sql
banderous: Can we write an integration test for these migrations? We should be able to write a test that runs the migrations up to this one, populates some data and then runs the last two.

lgonczar:
I've removed the last migration from this pull request, so that it can be merged in quickly. Populating the column will be a part of another one, as it always should have.

banderous:
We want the creation and population of the columns to be atomic, so they should be part of the same migration script - we want the columns populated or no columns at all, which will happen if they're in the same script which is run in a transaction

lgonczar:
I'll add the update to that last script, then, as it hasn't been applied and won't be for a bit.

banderous:
Approved but we should follow up with integration tests for our migration.

PR #34, opened 2018-03-26
Expand SendLetter integration test to run scheduled tasks

banderous:
### Change description ###

Expand SendLetter test to run scheduled tasks and verify letters are posted

This is a first attempt at an end-to-end test that runs both UploadLetters & MarkLettersPosted tasks, through an expansion of the existing SendLetter test.

- Introduced new XeroxReportWriter class to generate Xerox CSV reports based on UUIDs.
- SendLetter test now does the following:
  - Submit SendLetter request
  - Start SFTP server
  - Run UploadLetters task
  - Examine SFTP and generate CSV reports for any PDFS uploaded
  - Run MarkLettersPosted task
  - Verify that the letter is marked as posted

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/util/XeroxReportWriter.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/util/XeroxReportWriter.java
@@ -3,6 +3,7 @@ package uk.gov.hmcts.reform.sendletter.util;
 import org.apache.commons.csv.CSVFormat;
 import org.apache.commons.csv.CSVPrinter;
 import org.assertj.core.util.Lists;
+import uk.gov.hmcts.reform.slc.services.steps.getpdf.FileNameHelper;
 
 import java.io.File;
 import java.io.FileWriter;
@@ -15,7 +16,7 @@ import java.util.stream.Stream;
 public final class XeroxReportWriter {
     public static void writeReport(Stream<UUID> letterIds, File reportFolder) throws IOException {
         // We expect these fields in Xerox reports
-        CSVFormat csvFileFormat = CSVFormat.EXCEL.withHeader(
+        CSVFormat csvFileFormat = CSVFormat.DEFAULT.withHeader(
             "Date",
             "Time",
             "Filename"
@@ -25,8 +26,7 @@ public final class XeroxReportWriter {
         List<List<String>> records = letterIds.map(x -> Lists.newArrayList(
             "2018-01-01",
             "10:30:53",
-            // Id is the third underscore separated part of filename.
-            String.format("a_b_%s", x)
+            FileNameHelper.generateName("aType", "aService", x, "pdf")
         )).collect(Collectors.toList());
 
         File report = new File(reportFolder, UUID.randomUUID() + ".csv");
--- a/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/util/XeroxReportWriter.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/util/XeroxReportWriter.java
@@ -1,0 +1,42 @@
+package uk.gov.hmcts.reform.sendletter.util;
+
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVPrinter;
+import org.assertj.core.util.Lists;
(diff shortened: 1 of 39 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -21,19 +22,19 @@
...
     public static String SMOKE_TEST_LETTER_TYPE = "smoke_test";
 
     private final LetterRepository repo;
-    private final Zipper zipper;
+    private final Zipper zipper = new Zipper();
     private final FtpClient ftp;
 
     @Autowired
-    public UploadLettersTask(LetterRepository repo, Zipper zipper, FtpClient ftp) {
+    public UploadLettersTask(LetterRepository repo, FtpClient ftp) {
```
rkondratowicz: why?

banderous: The Zipper was causing the Application context to fail to initialise (though that may have been fixed).

The Zipper isn't currently mocked/asserted on in any tests and doesn't have any state, so we could replace it with static method calls?

banderous: For example why is the FileNameHelper static methods but the Zipper an instance class?

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/util/XeroxReportWriter.java
```diff
@@ -0,0 +1,42 @@
...
+import java.io.IOException;
+import java.util.List;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public final class XeroxReportWriter {
+    public static void writeReport(Stream<UUID> letterIds, File reportFolder) throws IOException {
+        // We expect these fields in Xerox reports
+        CSVFormat csvFileFormat = CSVFormat.EXCEL.withHeader(
```
rkondratowicz: parser uses `CSVFormat.DEFAULT` I think

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/util/XeroxReportWriter.java
```diff
@@ -0,0 +1,42 @@
...
+            "Time",
+            "Filename"
+        );
+
+        // Prepare a record for each of our letters.
+        List<List<String>> records = letterIds.map(x -> Lists.newArrayList(
+            "2018-01-01",
+            "10:30:53",
+            // Id is the third underscore separated part of filename.
+            String.format("a_b_%s", x)
```
rkondratowicz: it would be safer to use existing `FilenameHelper` to generat this

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
banderous: will fix

src/integrationTest/resources/application.yaml
```diff
@@ -0,0 +1,40 @@
...
+  smoke-test-target-folder: ${FTP_SMOKE_TEST_TARGET_FOLDER:SMOKE_TEST}
+  reports-folder: reports
+  reports-cron: ${FTP_REPORTS_CRON:0 30 0 * * *}
+  fingerprint: SHA1:2Fo8c/96zv32xc8GZWbOGYOlRak=
+  username: irrelevant
+  privateKey: ${FTP_PRIVATE_KEY}
+  publicKey: ${FTP_PUBLIC_KEY}
+  downtime:
+    from: ${FTP_DOWNTIME_FROM:17:00}
+    to: ${FTP_DOWNTIME_TO:18:00}
```
rkondratowicz: we should be able to remove those environment variables here if we don't use this functionality in tests

banderous: These values are needed for the ApplicationContext to load at present.

In a follow up PR I will set these to real values and we will ensure the scheduled tasks execute.

PR #42, opened 2018-03-28
Add unit test for FtpAvailabilityChecker

KarolJastrzebski:

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/test/java/uk/gov/hmcts/reform/sendletter/services/FtpAvailabilityCheckerTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/sendletter/services/FtpAvailabilityCheckerTest.java
@@ -30,6 +30,36 @@ public class FtpAvailabilityCheckerTest {
         assertThatFtpCheck("06:00", "08:00", "07:00").isFalse();
     }
 
+    @Test
+    public void ftp_is_unavailable_exactly_at_downtime_start() {
+        assertThatFtpCheck("17:30", "20:00", "17:30").isFalse();
+    }
+
+    @Test
+    public void ftp_is_unavailable_exactly_at_downtime_end() {
+        assertThatFtpCheck("17:30", "20:00", "20:00").isFalse();
+    }
+
+    @Test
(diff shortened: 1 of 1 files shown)


src/test/java/uk/gov/hmcts/reform/sendletter/services/FtpAvailabilityCheckerTest.java
```diff
@@ -0,0 +1,47 @@
...
+import org.junit.Test;
+
+import java.time.LocalTime;
+import java.time.format.DateTimeParseException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class FtpAvailabilityCheckerTest {
+
+    private static AbstractBooleanAssert<?> assertThatFtpCheck(String fromHour, String toHour, String now) {
```
lgonczar: It would be best to have this at the bottom, after test methods.

src/test/java/uk/gov/hmcts/reform/sendletter/services/FtpAvailabilityCheckerTest.java
```diff
@@ -0,0 +1,47 @@
...
+        assertThatFtpCheck("16:00", "17:00", "15:00").isTrue();
+    }
+
+    @Test
+    public void ftp_is_available_after_downtime() {
+        assertThatFtpCheck("08:00", "09:00", "10:00").isTrue();
+    }
+
+    @Test
+    public void ftp_is_unavailable_during_downtime() {
```
lgonczar: Please add edge cases from the existing tests (consumer).

src/test/java/uk/gov/hmcts/reform/sendletter/services/FtpAvailabilityCheckerTest.java
KarolJastrzebski: Although private, it is a static method and should be easier to spot at the beginning of the class. I would keep it where it is, but could move if you really, really wanted. :)

PR #43, opened 2018-03-28
Feature: Short circuit FTP upload on failure

doncem:
~Introducing `FailedToUpload` status.~ Keep original records unchanged in case of failure. Break the upload process in case failed to upload file

- fix exception catching in upload task
- move ftp exception to general exceptions for new package
- cover all scenarios for upload task

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -65,12 +66,18 @@ public void run() {
 
                 // remove pdf content, as it's no longer needed
                 letter.setPdf(null);
+            } catch (FtpException exception) {
+                letter.setState(LetterState.FailedToUpload);
 
+                logger.error(String.format("Exception uploading letter %s", letter.getId()), exception);
+            } catch (IOException exception) {
+                logger.error(String.format("Failed to zip document for letter %s", letter.getId()), exception);
```
rkondratowicz: how do you know it's zipping?

doncem: by looking into code. zipper throws IOException

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -65,12 +66,18 @@ public void run() {
 
                 // remove pdf content, as it's no longer needed
                 letter.setPdf(null);
+            } catch (FtpException exception) {
+                letter.setState(LetterState.FailedToUpload);
```
rkondratowicz: so we'll never try to upload it again?

doncem: There can be other task to loop through failed letters as it is specific scenario. If problem is continuous we assuring to not keep on hitting with same (most likely bigger) volume of documents.
This is as per suggestion in previous PR plus functionality kind of ported from producer/consumer repos

src/test/java/uk/gov/hmcts/reform/sendletter/services/FtpUploadTest.java
```diff
@@ -9,21 +9,22 @@
...
 public class FtpUploadTest {
 
     @Test
     public void connects_to_ftp() throws Exception {
-        try (LocalSftpServer server = LocalSftpServer.create()) {
-            FtpHelper.getClient(server.port).testConnection();
-        }
+        LocalSftpServer server = LocalSftpServer.create();
+        FtpHelper.getSuccessfulClient(port).testConnection();
+        server.close();
```
rkondratowicz: why not try-with-resource?

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterState.java
```diff
@@ -3,5 +3,6 @@
 public enum LetterState {
     Created,
     Uploaded,
+    FailedToUpload,
```
banderous: Why are we creating this state? If an upload fails it should be left in the Created state where it will be retried, there should not be a state for this.

We should mark letters as stale that have been uploaded to Xerox more than X days ago.

src/test/java/uk/gov/hmcts/reform/sendletter/services/FtpUploadTest.java
doncem: there's nothing to try here plus `server` would be unused code

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterState.java
banderous: Having discussed with Rafal, we think it would be best not to have any new state but instead just have this job generate alerts for letters that are in the Uploaded state more than a threshold time - thoughts?

src/test/java/uk/gov/hmcts/reform/sendletter/services/FtpUploadTest.java
banderous: But then if there's an exception in testConnection the server will not be closed.

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterState.java
doncem: As to why question, same [answer](https://github.com/hmcts/send-letter-service/pull/43#discussion_r177739809) can be applied. As to reporting the stale letters (which is being an `Uploaded` state + threshold) - next pr pending to do that exactly.

On top of first question I can add that not only functionality is kind of ported from producer/consumer, but also adapted to new column `state` which is highly important in the process. Moreover making `is_failed` attribute irrelevant. In case of a need to implement retry logic it has to come through the sprint planning with agreed specs rather than speculating and giving hypothetical suggestions to amend the agreed functionality.

That being said, suggestions are great 👍. Main point of not allowing retry for now is answered earlier - not ram the ftp which continues failing with same documents generating unnecessary volumes.

src/test/java/uk/gov/hmcts/reform/sendletter/services/FtpUploadTest.java
doncem: checkstyle and pmd doesn't give warning for unused code - rolling back 👍

banderous:
Can a document be in a terminal state where it cannot be uploaded? Why
would we expect it to be impossible to upload a specific letter?

Failing to retry uploading a letter is a bug rather than a feature; we
expect letters to fail to upload in the normal course of operation.

I agree that we don't want to hammer the FTP site though, we should
probably adjust our UploadLetters task to abort on the first IOException.

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
```diff
@@ -79,4 +91,62 @@ public void uploads_file_to_sftp_and_sets_letter_status_to_uploaded() throws Exc
             assertThat(l.getPdf()).isNull();
         }
     }
+
+    @Test
+    public void should_fail_to_upload_to_sftp_and_set_letter_status_as_failed() throws Exception {
```
rkondratowicz: needs updating as well

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
```diff
@@ -79,4 +91,62 @@ public void uploads_file_to_sftp_and_sets_letter_status_to_uploaded() throws Exc
...
+        LetterService s = new LetterService(repository, new ObjectMapper());
+        UUID id = s.send(SampleData.letter(), "service");
+
+        // Invoke the upload job.
+        try (LocalSftpServer server = LocalSftpServer.create()) {
+            UploadLettersTask task = new UploadLettersTask(
+                repository,
+                new Zipper(),
+                FtpHelper.getFailingClient(LocalSftpServer.port),
```
rkondratowicz: this could be moved to 'given' stage of the test

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
doncem: 😇

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -68,9 +69,13 @@ public void run() {
...
-                logger.debug("Marked letter {} as Uploaded", letter.getId());
-            } catch (Exception e) {
-                logger.error("Exception uploading letter {}", letter.getId(), e);
+                logger.debug("Marked letter {} as {}", letter.getId(), letter.getStatus());
+            } catch (FtpException exception) {
+                logger.error("Exception uploading letter {}", letter.getId());
+
+                throw exception;
+            } catch (IOException exception) {
+                logger.error(String.format("Failed to zip document for letter %s", letter.getId()), exception);
```
rkondratowicz: we can't make this assumption. Either throw custom exception specific to zipping, or change the error message

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -68,14 +69,18 @@ public void run() {
...
                 repo.saveAndFlush(letter);
 
-                logger.debug("Marked letter {} as Uploaded", letter.getId());
-            } catch (Exception e) {
-                logger.error("Exception uploading letter {}", letter.getId(), e);
+                logger.debug("Marked letter {} as {}", letter.getId(), letter.getStatus());
+            } catch (FtpException exception) {
+                logger.error("Exception uploading letter {}", letter.getId());
+
+                throw exception;
```
banderous: I suggest we don't rethrow the FtpException but instead log it and break out of the foreach loop; if we rethrow the exception then we will roll back the transaction that may have uploaded some letters successfully and set them to the uploaded state.

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -68,14 +74,18 @@ public void run() {
...
                 repo.saveAndFlush(letter);
 
-                logger.debug("Marked letter {} as Uploaded", letter.getId());
-            } catch (Exception e) {
-                logger.error("Exception uploading letter {}", letter.getId(), e);
+                logger.debug("Marked letter {} as {}", letter.getId(), letter.getStatus());
+            } catch (FtpException exception) {
+                logger.error(String.format("Exception uploading letter %s", letter.getId()), exception);
+
+                break;
```
rkondratowicz: add a test

PR #52, opened 2018-03-29
Enable scheduling of tasks

banderous:
### Change description ###

This test adds Scheduling to the service according to the old schedule defined in the consumer.

Also adds a new end to end integration test that tests the scheduling, and removes an obsolete test from the SendLetterTest.

TODO in followup PR is make use of SerialTaskRunner to ensure tasks do not run concurrently.

```
[ ] Yes
[ ] No
```


src/integrationTest/java/uk/gov/hmcts/reform/sendletter/e2e/EndToEndTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Test
+    public void should_upload_letter_and_mark_posted() throws Exception {
+        try (LocalSftpServer server = LocalSftpServer.create()) {
+            send(readResource("letter.json"))
+                .andExpect(MockMvcResultMatchers.status().isOk())
+                .andReturn();
+
+            // Report should be uploaded
+            await().atMost(60, SECONDS).until(() -> tryCreateXeroxReport(server));
```
rkondratowicz: Why is csv created here (by the test code)?
Can it just be 'file appeared on sftp'?

src/integrationTest/resources/application.yaml
```diff
@@ -33,8 +33,44 @@ ftp:
...
   downtime:
     from: ${FTP_DOWNTIME_FROM:17:00}
     to: ${FTP_DOWNTIME_TO:18:00}
+
+# Don't run scheduled tasks unless
+# explicitly turned on by tests.
+scheduling.enabled: false
+
+tasks:
+  uploadInterval: ${UPLOAD_LETTERS_INTERVAL:1}
```
rkondratowicz: worth adding a comment on unit of time here imho

src/main/resources/application.yaml
```diff
@@ -58,3 +58,8 @@ ftp:
   downtime:
     from: ${FTP_DOWNTIME_FROM:17:00}
     to: ${FTP_DOWNTIME_TO:18:00}
+
+tasks:
+  uploadInterval: ${UPLOAD_LETTERS_INTERVAL:30000}
```
rkondratowicz: worth adding a comment on unit of time here

src/integrationTest/resources/application.yaml
```diff
@@ -33,8 +33,44 @@ ftp:
...
+              eC+chICSB8O4E+3z3Qk9jdu4WWbU52G4tX0v56Nm5d/beRyAhtyZR3vwB12EpPy9
+              2ULbw/qJfArOE5TWqMIHWt34ZvNm/S559Jj3GLOwk1wIB/pLbVh79LG+XYqzG47B
+              h9cix6+q+1qGlI7huSIz4QKBgBN0FPxxCDJDuuIas31CObCDeD8AzTMxlicIa+Dt
+              05YInkse/nu4e6hOzGiDWPnlt6DYr78Iio7uC1ILhvQL2RC+dcQaWQESSwOtOWai
+              LggZp441g3mW1HOiOToBeWesw/SgnEHntRgCIVQukjpOA1yvl/eY6Ebw0LJbj85m
+              MuohAoGAO8M64mpnVF9avuLGsv/yuX/g5/MUO/x0jZReLUmLnkpU4i5YAU2EJ8oD
+              C1ugH+oTYtMY0H4ytYPYeeEiNMyVYg3GE7U2dhpYoafRR1q6AkKr4E58Q+ThraIl
+              gH158Quekp3NyrTWuWo3OlbPwGvRvMPV0URAH9UX21mMkuqpBGE=
+              -----END RSA PRIVATE KEY-----
+  publicKey: ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDHomWdZ5Q2ZL/ElmDWRwoztvHaj9nr/l0TNiDsx1S3YIHbqcBRpffq4j/lsC8IhVthmGIiaTQU9e668otDow7KBG9vCyonBW4nx9VnLh3aEEzGWxLT9JfxNE4spRveH5SwsleTI+ljFEZtP0T5ha7YhJ3xtrVnPbXwN1BWsoV4omd7Mf4uj7mSA/gjC1gBu7GLwScuAeVp7S6WF2tYB3EbG+T99C0Y0NQAbji78Oe1hX9rjkj/fQt3Kv2OEKfcTXH3fXyp2q6LjDNyMiDylWQAqwxXOBk1swlL57v4K7mWtKSA6054yBJr+ATSQ+vG5FfsqNDQ9YbsT5gvMf8FWuFx alex@MacBook-Pro-7.local
```
doncem: 😯 . can we have _key_ specific values separately per resource file? then load them manually where needed? makes app config/properties sensible

src/main/resources/application.yaml
```diff
@@ -58,3 +58,8 @@ ftp:
...
+  reports-cron: ${FTP_REPORTS_CRON:0 30 0 * * *}
```
rkondratowicz: let's stick to one convention (`foo-bar` here and `fooBar` above)

src/main/resources/application.yaml
```diff
@@ -58,3 +58,8 @@ ftp:
...
+
```
rkondratowicz: missing `scheduling.enabled` config entry

src/integrationTest/resources/application.yaml
```diff
@@ -33,8 +33,44 @@ ftp:
...
     to: ${FTP_DOWNTIME_TO:18:00}
+
+# Don't run scheduled tasks unless
+# explicitly turned on by tests.
+scheduling.enabled: false
+
+tasks:
+  uploadInterval: ${UPLOAD_LETTERS_INTERVAL:1}
+  # Run every second.
+  reports-cron: ${FTP_REPORTS_CRON:*/1 * * * * *}
```
doncem: there's one in ftp scope ^

doncem: so do we camel case or dashing through the snow krhem words?

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
```diff
@@ -19,6 +22,8 @@
  * Fetches reports from Xerox SFTP concerning posted
  * letters and sets posted letters as Posted in the database.
  */
+@Component
+@ConditionalOnProperty(value = "scheduling.enabled", havingValue = "true", matchIfMissing = true)
```
doncem: boolean flags don't really make sense in checking whether has a value of `true`

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
```diff
@@ -38,6 +43,12 @@ public MarkLettersPostedTask(LetterRepository repo,
         this.parser = parser;
     }
 
+
+    @Scheduled(cron = "${tasks.reports-cron}")
```
doncem: perhaps best to be more specific and e.g. say it is `print-report-cron` or smth similar

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -24,6 +26,7 @@
 import static java.time.LocalDateTime.now;
 
 @Component
+@ConditionalOnProperty(value = "scheduling.enabled", havingValue = "true", matchIfMissing = true)
```
doncem: same with conditioning

src/main/resources/application.yaml
```diff
@@ -58,3 +58,8 @@ ftp:
   downtime:
     from: ${FTP_DOWNTIME_FROM:17:00}
     to: ${FTP_DOWNTIME_TO:18:00}
```
doncem: i reckon it's not the latest code, iirc, should be 16-17

src/main/resources/application.yaml
doncem: i guess question about camel case and dashes should've been in main app config. putting a reminder here

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/e2e/EndToEndTest.java
banderous: We need to generate the report so the MarkLettersPosted task processes it and marks the letter posted

src/integrationTest/resources/application.yaml
banderous: It would be nice to but how are these values set in production? My understanding is that this file is populated by terraform and the fields are filled in, so adding the fields to the file here mimics the production setup.

rkondratowicz: @banderous environment variable is set

build.gradle
```diff
@@ -85,6 +85,9 @@ task integration(type: Test, description: 'Runs the integration tests.', group:
   classpath = sourceSets.integrationTest.runtimeClasspath
 
   environment("APPINSIGHTS_INSTRUMENTATIONKEY", "test-key")
+  environment("FTP_PRIVATE_KEY", file('src/test/resources/keypair').text)
+  environment("FTP_PUBLIC_KEY", file('src/test/resources/keypair.pub').text)
```
doncem: should be in integration module rather than test

PR #57, opened 2018-03-29
cleanup config file

rkondratowicz:
Remove:
- duplicated FTP settings
- consumer name settings (no longer used)
- service bus ttl settings (no longer used)


src/main/resources/application.yaml
```diff
@@ -5,20 +5,10 @@ management:
...
-
-status-update-service-name: ${STATUS_UPDATE_SERVICE_NAME:sendletterconsumer}
-
 idam:
   s2s-auth:
     url: ${S2S_URL}
 
-ftp:
-  smoke-test-target-folder: ${FTP_SMOKE_TEST_TARGET_FOLDER:SMOKE_TEST}
-  downtime:
```
lgonczar: Why this?

lgonczar: Ok, duplicated

PR #60, opened 2018-03-29
Remove not used code

rkondratowicz:
consumer no longer calls this service

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/services/AuthService.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/services/AuthService.java
@@ -1,34 +1,16 @@
 package uk.gov.hmcts.reform.sendletter.services;
 
-import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/services/AuthService.java
```diff
@@ -1,49 +1,23 @@
...
 
 @Component
 public class AuthService {
 
-    /** Name of the service that can update letter status. **/
-    private final String statusAccessorService;
     private final AuthTokenValidator authTokenValidator;
 
-    public AuthService(
-        @Value("${status-update-service-name}") String statusAccessorService,
```
nitinprabhu: Do we need to remove this https://github.com/hmcts/send-letter-service/blob/master/src/integrationTest/resources/application.yaml#L1?

rkondratowicz: yup!

PR #62, opened 2018-03-29
Upgrade jacoco

rkondratowicz:
should not report on untested private constructors in utility classes now 🤞


rkondratowicz:
+2.91% 🎉

PR #63, opened 2018-04-02
Feature: stale letters

doncem:

[Migrating failed letter code to producer](https://tools.hmcts.net/jira/browse/RPE-320)

### Change description ###

- new stale letter task ~with projection class used in query~
- dynamically pick up cut off time from ftp availability checker

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/StaleLettersTask.java
```diff
@@ -0,0 +1,47 @@
...
+        AppInsights insights,
+        FtpAvailabilityChecker checker
+    ) {
+        this.repo = repo;
+        this.insights = insights;
+        this.staleCutOffTime = checker.getDowntimeStart();
+    }
+
+    public void run() {
+        logger.trace("Running job");
```
nitinprabhu: I think this should be info?

doncem: ah, changes after lot of PRs got merged last week. this one was mostly rebase and squash the old noise

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
```diff
@@ -2,18 +2,32 @@
...
 
+import java.sql.Timestamp;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.Stream;
 import javax.persistence.LockModeType;
 
 public interface LetterRepository extends JpaRepository<Letter, UUID> {
+
+    String STALE_LETTERS_QUERY = "select"
```
nitinprabhu: could be better if we could store in some external file similar to https://github.com/spring-projects/spring-data-jpa/blob/master/src/test/resources/META-INF/jpa-named-queries.properties

src/test/java/uk/gov/hmcts/reform/sendletter/logging/AppInsightsTest.java
```diff
@@ -109,6 +120,36 @@ public void should_track_message_acknowledgement_event_for_fail_case() {
         verifyNoMoreInteractions(telemetry);
     }
 
+    @Test
+    public void should_track_event_of_not_printed_letter() {
+        StaleLetter letter = mock(StaleLetter.class);
```
nitinprabhu: May be you could create an actual object here rather than mock and then in below verification also assert the values?

src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
```diff
@@ -43,6 +50,20 @@ public void trackFtpUpload(java.time.Duration duration, boolean success) {
...
 
+    public void trackStaleLetter(StaleLetter staleLetter) {
+        LocalDateTime sentToPrint = LocalDateTime.ofInstant(staleLetter.getSentToPrintAt().toInstant(), ZoneOffset.UTC);
+        Map<String, String> properties = new HashMap<>();
+
+        properties.put("letterId", staleLetter.getId().toString());
+        properties.put("messageId", staleLetter.getMessageId());
+        properties.put("service", staleLetter.getService());
+        properties.put("type", staleLetter.getType());
+        properties.put("weekday", sentToPrint.getDayOfWeek().name());
```
nitinprabhu: may be better variable name here e.g `sentToPrintDayOfWeek`?

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/StaleLettersTaskTest.java
```diff
@@ -0,0 +1,140 @@
...
+        task.run();
+
+        // then
+        verify(insights, never()).trackStaleLetter(any(StaleLetter.class));
+    }
+
+    private Letter createLetter(LocalTime withSentToPrintAt) {
+        Letter letter = new Letter(MessageIdProvider.randomMessageId(), "service", null, "type", null);
+
+        if (withSentToPrintAt != null) {
```
nitinprabhu: we could use `sentToPrintAtTime` instead of `withSentToPrintAt`?

src/test/java/uk/gov/hmcts/reform/sendletter/logging/AppInsightsTest.java
doncem: remnant of the past. will update test

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/StaleLettersTask.java
```diff
@@ -0,0 +1,47 @@
...
+
+        Timestamp staleCutOff = Timestamp.valueOf(
```
nitinprabhu: could you please log the cutoffTime as it might be useful while debugging?

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
banderous: If the objective is to avoid loading the PDF column then instead of having an explicit query we should be able to make use of lazy loading for the pdf column using the [Lob annotation](https://docs.jboss.org/hibernate/jpa/2.1/api/javax/persistence/Lob.html) on the pdf field.

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/StaleLettersTaskTest.java
```diff
@@ -0,0 +1,140 @@
...
+        // then
+        verify(insights).trackStaleLetter(captor.capture());
+
+        // and
+        assertThat(captor.getAllValues()).hasSize(1);
+        assertThat(captor.getValue().getId()).isEqualTo(letter.getId());
+    }
+
+    @Test
+    public void should_not_pick_up_letter_if_sent_to_print_happened_from_the_cutoff() {
```
rkondratowicz: something wrong in this sentence

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/StaleLettersTask.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Autowired
+    public StaleLettersTask(
+        LetterRepository repo,
+        AppInsights insights,
+        FtpAvailabilityChecker checker
+    ) {
+        this.repo = repo;
+        this.insights = insights;
+        this.staleCutOffTime = checker.getDowntimeStart();
```
rkondratowicz: can we just inject the actual time instead?

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
doncem: `Lob` is not needed, apparently. It maps to different column type, `Basic` usage is enough: https://stackoverflow.com/a/3739047/984221

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/StaleLettersTask.java
doncem: tempting suggestion. is it good idea to have configuration values passed in different places though?

src/main/java/uk/gov/hmcts/reform/sendletter/entity/StaleLetter.java
```diff
@@ -3,49 +3,15 @@
 import java.util.Date;
 import java.util.UUID;
 
-public class StaleLetter {
+public interface StaleLetter {
```
nitinprabhu: Since we are using projection I would recommend you rename this class as `StaleLettersOnly` as it would be very clear then rather than sounding like a POJO.

src/main/java/uk/gov/hmcts/reform/sendletter/entity/Letter.java
```diff
@@ -43,6 +46,8 @@
     private String type;
     @Enumerated(EnumType.STRING)
     private LetterStatus status = LetterStatus.Created;
+
+    @Basic(fetch = LAZY)
     private byte[] pdf;
```
rkondratowicz: care to explain?

doncem: all hibernate queries are eagerly loaded by default. to avoid fetching pdf whenever we query for `Letter` - set `basic` usage of a resource fetch

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/StaleLettersTaskTest.java
```diff
@@ -0,0 +1,141 @@
...
+
+        // then
+        verify(insights, never()).trackStaleLetter(any(Letter.class));
+    }
+
+    @Test
+    @SuppressWarnings("VariableDeclarationUsageDistance")
+    public void should_report_to_insights_when_there_is_an_unprinted_letter() {
+        // given
+        Letter letter = createLetter(secondBeforeCutOff);
```
rkondratowicz: looks like it's the date when letter was created, but it's when it was sent to print. worth updating method name imgo

PR #70, opened 2018-04-03
Update java logging

doncem:
Notes:

- use latest version of java logging
- include application version for app insights
- use exception helper to correctly implement requirement

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/sendletter/exception/DocumentZipException.java
```diff
@@ -1,8 +1,15 @@
...
 
-public class DocumentZipException extends RuntimeException {
+import uk.gov.hmcts.reform.logging.exception.AlertLevel;
+import uk.gov.hmcts.reform.logging.exception.UnknownErrorCodeException;
+
+/**
+ * SonarQube reports as error. Max allowed - 5 parents
+ */
+@SuppressWarnings("squid:MaximumInheritanceDepth")
+public class DocumentZipException extends UnknownErrorCodeException {
```
banderous: I'm not familiar with reform logging, what is UnknownErrorCodeException? Seems like an odd name for a common base class.

src/main/java/uk/gov/hmcts/reform/sendletter/exception/LetterNotFoundException.java
```diff
@@ -1,11 +1,18 @@
...
-public class LetterNotFoundException extends RuntimeException {
+/**
+ * SonarQube reports as error. Max allowed - 5 parents
+ */
+@SuppressWarnings("squid:MaximumInheritanceDepth")
+public class LetterNotFoundException extends UnknownErrorCodeException {
 
     public LetterNotFoundException(String letterId, Throwable cause) {
-        super("Letter with ID '" + letterId + "' not found", cause);
+        super(AlertLevel.P4, "Letter with ID '" + letterId + "' not found", cause);
```
banderous: Why is this P4 compared to P1 for the other exceptions?

src/main/java/uk/gov/hmcts/reform/sendletter/exception/DocumentZipException.java
banderous: Having taken a look at java-logging, would it make sense to merge UnknownErrorCodeException into the AbstractLoggingException base class, since UnknownErrorCodeException has no fields of its own and only provides constructor overloads? That might let us get rid of the SuppressWarnings annotations.

We could still keep UnknownErrorCodeException as a type to preserve backwards compatibility and avoid breaking clients, but modify our code to derive from AbstractLoggingException.

src/main/java/uk/gov/hmcts/reform/sendletter/exception/LetterNotFoundException.java
doncem: `LetterNotFound` is kind of customer error and it is the least priority support should care about hence P4

src/main/java/uk/gov/hmcts/reform/sendletter/exception/DocumentZipException.java
doncem: `UnknownErrorCodeException` was introduced as a quick _unduplication_ of error code requirement in the abstract constructors. both fields are hmcts requirement so we quickly sketched former and found out about inheritance depth. Considered to investigate this in a more calm period and/or during BAU whenever one of us remembers about it and has time, priority.
Regardless, it will have to become another java-logging version bump and fix cascaded.

PR #74, opened 2018-04-03
zip pdf before saving to db

rkondratowicz:
Note that no migration is needed as this service has not yet been deployed


src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
```diff
@@ -117,12 +101,8 @@ public void should_fail_to_upload_to_sftp_and_stop_from_uploading_any_other_lett
             task.run();
 
             // then
-            // verify zipper was called only once
-            verify(fakeZipper).zip(anyString(), any(PdfDoc.class));
-
             // file does not exist in SFTP site.
-            File[] files = server.lettersFolder.listFiles();
-            assertThat(files.length).isEqualTo(0);
+            assertThat(server.lettersFolder.listFiles().length).isEqualTo(0);
```
banderous: could do `assertThat(server.lettersFolder.list()).isEmpty()`

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/services/LetterServiceTest.java
```diff
@@ -57,20 +56,19 @@ public void tearDown() {
...
         Letter result = letterRepository.findOne(id);
-        DataSource dataSource = new ByteArrayDataSource(new ByteArrayInputStream(result.getFileContent()));
-        PreflightParser pdfParser = new PreflightParser(dataSource);
-        pdfParser.parse();
-        PreflightDocument document = pdfParser.getPreflightDocument();
-        // This will throw an exception if the file format is invalid,
-        // but ignores more minor errors such as missing metadata.
-        document.validate();
+
+        ZipInputStream zip = new ZipInputStream(new ByteArrayInputStream(result.getFileContent()));
```
banderous: We could keep the existing checks and validate that it unzips to a valid pdf, just by wrapping the ByteArrayDataSource constructor parameter in a ZipInputStream.

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/services/LetterServiceTest.java
```diff
@@ -57,20 +56,19 @@ public void tearDown() {
     }
 
     @Test
-    public void generates_and_saves_pdf() throws IOException {
+    public void generates_and_saves_zip_file() throws IOException {
```
banderous: generates_and_saves_zipped_pdf if we follow the comment below

banderous:
It might be worth updating the end to end test to validate that a valid, zipped pdf is uploaded to the local SFTP server, to ensure we aren't double zipping the content (I don't believe we are, but I had to study the code for a while to figure that out)

PR #76, opened 2018-04-03
Accept request with new letter model

rkondratowicz:
Using media type versioning (see https://hmcts.github.io/restful-api-standards/#114).

This PR deals with calling the correct service method only.

Old endpoint supports both new type and default `application/json`.


src/main/java/uk/gov/hmcts/reform/sendletter/services/LetterService.java
```diff
@@ -59,6 +61,11 @@ public UUID send(LetterRequest letter, String serviceName) {
             .orElseGet(() -> saveNewLetterAndReturnId(letter, messageId, serviceName));
     }
 
+    @Transactional
+    public UUID send(LetterWithPdfsRequest letter, String serviceName) {
+        throw new NotImplementedException();
+    }
```
rkondratowicz: will be added in next PR

src/test/java/uk/gov/hmcts/reform/sendletter/controllers/SendLetterControllerTest.java
```diff
@@ -121,11 +125,35 @@ public void should_return_400_client_error_when_letter_is_with_more_than_10_docu
...
+        for (String type : supportedContentTypes) {
+            mockMvc
+                .perform(
+                    post("/letters")
+                        .contentType(type)
+                        .header("ServiceAuthorization", "auth-header-value")
+                        .content(readResource("controller/letter/v1/letter.json"))
+                )
+                .andExpect(status().isOk());
+        }
```
rkondratowicz: can't use `.foreach`, sorry

PR #78, opened 2018-04-04
Load keypair files only when they exist

KarolJastrzebski:
Fix blocked gradle build.

```
[ ] Yes
[x] No
```


KarolJastrzebski:
Not necessary as it seems to be a path issue with gradle on linux.

PR #82, opened 2018-04-04
Add unit test for pdf creator

rkondratowicz:
Mostly copied from https://github.com/hmcts/send-letter-consumer-service/blob/master/src/test/java/uk/gov/hmcts/reform/slc/services/steps/getpdf/PdfCreatorTest.java
with slight changes:
- no calls to pdf-service api
- no calls to app insights

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/services/LetterService.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/services/LetterService.java
@@ -7,6 +7,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
+import uk.gov.hmcts.reform.pdf.generator.HTMLToPDFConverter;
 import uk.gov.hmcts.reform.sendletter.entity.Letter;
 import uk.gov.hmcts.reform.sendletter.entity.LetterRepository;
 import uk.gov.hmcts.reform.sendletter.exception.LetterNotFoundException;
@@ -33,7 +34,7 @@ public class LetterService {
 
     private static final Logger log = LoggerFactory.getLogger(LetterService.class);
 
-    private final PdfCreator pdfCreator = new PdfCreator(new DuplexPreparator());
+    private final PdfCreator pdfCreator = new PdfCreator(new DuplexPreparator(), new HTMLToPDFConverter()::convert);
     private final LetterRepository letterRepository;
     private final Zipper zipper;
     private final ObjectMapper mapper;
--- a/src/main/java/uk/gov/hmcts/reform/slc/config/PdfConversionConfig.java
+++ b/src/main/java/uk/gov/hmcts/reform/slc/config/PdfConversionConfig.java
@@ -10,7 +10,6 @@ public class PdfConversionConfig {
 
     @Bean
     public IHtmlToPdfConverter htmlToPdfConverter() {
-        HTMLToPDFConverter converter = new HTMLToPDFConverter();
-        return converter::convert;
+        return new HTMLToPDFConverter()::convert;
     }
 }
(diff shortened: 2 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/services/LetterService.java
```diff
@@ -33,7 +34,7 @@
 
     private static final Logger log = LoggerFactory.getLogger(LetterService.class);
 
-    private final PdfCreator pdfCreator = new PdfCreator(new DuplexPreparator());
+    private final PdfCreator pdfCreator = new PdfCreator(new DuplexPreparator(), new HTMLToPDFConverter()::convert);
```
KarolJastrzebski: This is already returned in PdfConversionConfig. Why not to use it?

rkondratowicz:
not everything at once

doncem:
yea. but you copied new files over :)

src/main/java/uk/gov/hmcts/reform/slc/config/PdfConversionConfig.java
```diff
@@ -0,0 +1,16 @@
+package uk.gov.hmcts.reform.slc.config;
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import uk.gov.hmcts.reform.pdf.generator.HTMLToPDFConverter;
+import uk.gov.hmcts.reform.slc.services.steps.getpdf.IHtmlToPdfConverter;
+
+@Configuration
+public class PdfConversionConfig {
```
KarolJastrzebski: Why not to take the burden of calling `new` throughout the app and return configured instance of PdfCreator instead?

src/main/java/uk/gov/hmcts/reform/slc/services/steps/getpdf/PdfCreator.java
```diff
@@ -15,10 +14,11 @@
 public class PdfCreator {
 
     private final DuplexPreparator duplexPreparator;
-    private static HTMLToPDFConverter converter = new HTMLToPDFConverter();
+    private final IHtmlToPdfConverter converter;
```
KarolJastrzebski: :+1:

src/main/java/uk/gov/hmcts/reform/slc/services/steps/getpdf/IHtmlToPdfConverter.java
```diff
@@ -0,0 +1,8 @@
+package uk.gov.hmcts.reform.slc.services.steps.getpdf;
+
+import java.util.Map;
+import java.util.function.BiFunction;
+
+public interface IHtmlToPdfConverter extends BiFunction<byte[], Map<String, Object>, byte[]> {
```
KarolJastrzebski: Friendly alias for lambda instance has its benefits. Alternatively, to keep inheritance hierarchy flat and open to other implementations, there are options like:

1.
```java

interface IHtmlToPdfConverter {
    byte[] convert(byte[] template, Map<String, Object> values);
}

class HtmlToPdfFuncAdapter implements IHtmlToPdfConverter {

    private final BiFunction<byte[], Map<String, Object>, byte[]> function;

    HtmlToPdfFuncAdapter(BiFunction<byte[], Map<String, Object>, byte[]> function) {
        this.function = function;
    }

    @Override
    public byte[] convert(byte[] template, Map<String, Object> values) {
        return function.apply(template, values);
    }
}
```
2. Use `BiFunction<byte[], Map<String, Object>, byte[]>` as it is quite technical and descriptive.

src/main/java/uk/gov/hmcts/reform/slc/config/PdfConversionConfig.java
rkondratowicz: `PdfCreator` is a `@Service` and is injected by Spring.

PR #85, opened 2018-04-04
Generate pdf from new letter model

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/slc/services/steps/getpdf/PdfCreator.java
```diff
@@ -34,9 +37,30 @@ public PdfCreator(DuplexPreparator duplexPreparator, IHtmlToPdfConverter convert
         return PdfMerger.mergeDocuments(docs);
     }
 
+    public byte[] create(LetterWithPdfsRequest letter) {
```
banderous: If all we need is the documents from the letter then we could take the documents as a parameter

src/main/java/uk/gov/hmcts/reform/slc/services/steps/getpdf/PdfCreator.java
```diff
@@ -34,9 +37,30 @@ public PdfCreator(DuplexPreparator duplexPreparator, IHtmlToPdfConverter convert
...
+        Asserts.notNull(letter, "letter");
+
+        List<byte[]> docs =
+            letter.documents
+                .stream()
+                .map(this::decodePdf)
+                .map(duplexPreparator::prepare)
```
banderous: There's some duplication here with the other method, we could perhaps have a helper method that calls the duplexPreparator and merges the docs.

src/main/java/uk/gov/hmcts/reform/slc/services/steps/getpdf/PdfCreator.java
banderous: We could change that method too

PR #88, opened 2018-04-04
Catch SQL exception in task runner

doncem:
~Don't throw exception to break the process.~ Log ~as~ error when task runner experienced SQLException. Solve the checked exception throughout the method chain

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
```diff
@@ -55,6 +56,11 @@ public void tryRun(int id, Runnable runnable) throws SQLException {
                     }
                 }
             }
+        } catch (SQLException exception) {
+            log.error(
+                String.format("SQL error occurred during task %s run", Task.getTaskFromLockId(id)),
+                new TaskRunnerException(exception)
```
rkondratowicz: why not `exception`?

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
```diff
@@ -55,6 +56,11 @@ public void tryRun(int id, Runnable runnable) throws SQLException {
                     }
                 }
             }
+        } catch (SQLException exception) {
+            log.error(
+                String.format("SQL error occurred during task %s run", Task.getTaskFromLockId(id)),
```
banderous: Let's just pass the Task enum in here instead of ints so we don't need to do this lookup

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
doncem: Because hmcts loggers require alert levels to be implemented

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/Task.java
```diff
@@ -3,9 +3,20 @@
 public enum Task {
     UploadLetters,
     MarkLettersPosted,
-    StaleLetters;
+    StaleLetters,
+    Unknown;
```
banderous: Can remove this if we pass the task enum to the SerialTaskRunner.

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
doncem: wanted to do so but was in fear in tempering existing implementation

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
```diff
@@ -55,6 +58,11 @@ public void tryRun(int id, Runnable runnable) throws SQLException {
                     }
                 }
             }
+        } catch (SQLException exception) {
+            log.error(
+                String.format("SQL error occurred during task %s run", task.name()),
+                new TaskRunnerException(exception)
```
rkondratowicz: I think it would be better to throw the new runtime exception you've created here instead of 'silently' failing. The whatever is calling this would handle exception in a desired way.

In the spirit of 'throw early, catch late' :)

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
```diff
@@ -55,6 +58,12 @@ public void tryRun(int id, Runnable runnable) throws SQLException {
                     }
                 }
             }
+        } catch (SQLException exc) {
+            TaskRunnerException exception = new TaskRunnerException(exc);
+
+            log.error(String.format("SQL error occurred during task %s run", task.name()), exception);
```
rkondratowicz: log caught exception

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
```diff
@@ -34,8 +35,10 @@ private SerialTaskRunner(DataSource source) {
      * No error is thrown if such a task is already running, the
      * supplied Runnable is simply not executed.
      */
-    public void tryRun(int id, Runnable runnable) throws SQLException {
+    void tryRun(Task task, Runnable runnable) {
+        int id = task.getLockId();
         log.info("Trying to lock {}", id);
```
banderous: We could also log the task enum as a string now that we have it available

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
```diff
@@ -34,27 +35,39 @@ private SerialTaskRunner(DataSource source) {
...
             }
+        } catch (SQLException exc) {
+            TaskRunnerException exception = new TaskRunnerException(exc);
+
+            log.error(
+                String.format("SQL error occurred during task %s run. Cause message: %s", name, exc.getMessage()),
+                exception
+            );
+
+            throw exception;
```
rkondratowicz: either log original exception here, or just throw custom one here and log in caller

PR #91, opened 2018-04-04
Prepare for handling letters with embedded pdfs

rkondratowicz:
Refactoring existing service so that handling both models will be easy


src/main/java/uk/gov/hmcts/reform/sendletter/services/LetterService.java
```diff
@@ -47,49 +48,48 @@ public LetterService(LetterRepository letterRepository, Zipper zipper, ObjectMap
...
-                UUID id = l.getId();
-                log.info("Same message found already created. Returning letter id {} instead", id);
-                return id;
-            })
-            .orElseGet(() -> saveNewLetterAndReturnId(letter, messageId, serviceName));
     }
 
     @Transactional
     public UUID send(LetterWithPdfsRequest letter, String serviceName) {
         throw new NotImplementedException();
```
rkondratowicz: this will simply be changed to
```
return send(
    new LetterInfo(generateChecksum(letter), serviceName, letter.type, letter.additionalData),
    () -> pdfCreator.createFromBase64Pdfs(letter.documents)
);
```

src/main/java/uk/gov/hmcts/reform/sendletter/services/LetterChecksumGenerator.java
```diff
@@ -10,7 +9,7 @@
     private LetterChecksumGenerator() {
     }
 
-    public static String generateChecksum(LetterRequest letter) {
+    public static String generateChecksum(Object letter) {
```
doncem: Should it be `Serializable` instead? Not strong about it since `serialize` has `Object` signature

src/main/java/uk/gov/hmcts/reform/sendletter/services/LetterService.java
```diff
@@ -47,49 +48,48 @@ public LetterService(LetterRepository letterRepository, Zipper zipper, ObjectMap
 
     @Transactional
```
doncem: Looks like both methods are wrappers/helpers of generic `send(LetterInfo, Supplier)` method. Maybe worth just moving `@Transactional` to that method instead?

src/main/java/uk/gov/hmcts/reform/sendletter/services/LetterService.java
doncem: Looks pretty much the same. Can review `LetterInfo` usage. But I suggest for later once current changes get into scope to avoid premature refactor

src/main/java/uk/gov/hmcts/reform/sendletter/services/LetterService.java
```diff
@@ -47,49 +48,48 @@ public LetterService(LetterRepository letterRepository, Zipper zipper, ObjectMap
...
+    }
+
+    private UUID saveNewLetter(LetterInfo letterInfo, byte[] pdfContent) {
         UUID id = UUID.randomUUID();
 
         byte[] zipContent = zipper.zip(
             new PdfDoc(
-                FileNameHelper.generateName(letterRequest.type, serviceName, id, "pdf"),
-                pdfCreator.createFromTemplates(letterRequest.documents)
+                FileNameHelper.generateName(letterInfo.type, letterInfo.serviceName, id, "pdf"),
```
doncem: For simplification filename could be also moved to `LetterInfo`: letter specific filename anyway; would make easy to scan code

src/main/java/uk/gov/hmcts/reform/sendletter/services/LetterService.java
```diff
@@ -47,49 +48,48 @@ public LetterService(LetterRepository letterRepository, Zipper zipper, ObjectMap
...
-            messageId,
-            serviceName,
-            mapper.valueToTree(letterRequest.additionalData),
-            letterRequest.type,
+            letterInfo.messageId,
+            letterInfo.serviceName,
+            mapper.valueToTree(letterInfo.additionalData),
+            letterInfo.type,
             zipContent
         );
```
doncem: Maybe to `LetterInfo` too? It is a helper class after all, should make use of it and simplify this crud operation

PR #93, opened 2018-04-05
Fix logged event in response exception handler

doncem:
Report log events to AppInsights with correct implementation avoiding "wrong implementation" errors magically appearing.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/ResponseExceptionHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/ResponseExceptionHandler.java
@@ -14,6 +14,8 @@ import org.springframework.web.bind.annotation.InitBinder;
 import org.springframework.web.context.request.WebRequest;
 import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
 import uk.gov.hmcts.reform.authorisation.exceptions.InvalidTokenException;
+import uk.gov.hmcts.reform.logging.exception.AbstractLoggingException;
+import uk.gov.hmcts.reform.sendletter.exception.InternalServerException;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/exception/UnauthenticatedException.java
```diff
@@ -11,4 +11,13 @@
     public UnauthenticatedException(String message) {
```
nitinprabhu: Do we need this constructor or can we remove it ?

doncem: This is used for missing header

src/main/java/uk/gov/hmcts/reform/sendletter/ResponseExceptionHandler.java
```diff
@@ -59,7 +61,7 @@ protected void activateDirectFieldAccess(DataBinder dataBinder) {
 
     @ExceptionHandler(InvalidTokenException.class)
     protected ResponseEntity<Void> handleInvalidTokenException(InvalidTokenException exc) {
-        log.warn(exc.getMessage(), exc);
+        log.warn(exc.getMessage(), new UnauthenticatedException(exc));
```
rkondratowicz: can we make `InvalidTokenException` extend the required base class instead?

doncem: Comes from library

rkondratowicz: let's change it :)

doncem: 🎉

PR #96, opened 2018-04-05
Simplify handling new letter model in service layer

rkondratowicz:
A separate method in service layer accepting the new model would require duplicating many tests.
This PR adds an interface that both models implement and leaves only one method in the letter service.

The only difference between two models is how document/pdf data is represented.
Service decides which `PdfCreator` method to call based on the type of model.


src/integrationTest/java/uk/gov/hmcts/reform/sendletter/entity/LetterTest.java
```diff
@@ -46,7 +46,7 @@ public void compatible_with_existing_records() throws Exception {
         ObjectMapper objectMapper = new ObjectMapper();
         uk.gov.hmcts.reform.sendletter.data.LetterRepository repo =
             new uk.gov.hmcts.reform.sendletter.data.LetterRepository(jdbcTemplate, objectMapper);
-        DbLetter dbLetter = new DbLetter(UUID.randomUUID(), "cmc", SampleData.letter());
+        DbLetter dbLetter = new DbLetter(UUID.randomUUID(), "cmc", SampleData.letterRequest());
```
rkondratowicz: added new method for sample new model, so renaming for clarity.

src/main/java/uk/gov/hmcts/reform/sendletter/services/LetterService.java
```diff
@@ -35,71 +35,91 @@
 
     private static final Logger log = LoggerFactory.getLogger(LetterService.class);
 
-    private final PdfCreator pdfCreator = new PdfCreator(new DuplexPreparator(), new HTMLToPDFConverter()::convert);
+    private final PdfCreator pdfCreator;
```
rkondratowicz: how did this slip through... :/

src/main/java/uk/gov/hmcts/reform/sendletter/services/LetterService.java
```diff
@@ -35,71 +35,91 @@
...
+    public LetterService(
+        LetterRepository letterRepository,
+        Zipper zipper,
+        ObjectMapper mapper
+    ) {
+        this(
+            new PdfCreator(new DuplexPreparator(), new HTMLToPDFConverter()::convert),
+            letterRepository,
+            zipper,
+            mapper
```
rkondratowicz: leaving this one so that e2e test, and other, pass (PdfCreator can't be autowired as it's in a separate package)

src/main/java/uk/gov/hmcts/reform/sendletter/exception/UnsupportedLetterRequestTypeException.java
```diff
@@ -0,0 +1,4 @@
+package uk.gov.hmcts.reform.sendletter.exception;
+
+public class UnsupportedLetterRequestTypeException extends RuntimeException {
```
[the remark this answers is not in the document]
rkondratowicz: This is actually server error.
Json payload must match model specified in the controller to get all the way to the service layer.

I'll add missing extend.

src/main/java/uk/gov/hmcts/reform/sendletter/exception/UnsupportedLetterRequestTypeException.java
```diff
@@ -0,0 +1,10 @@
+package uk.gov.hmcts.reform.sendletter.exception;
+
+import uk.gov.hmcts.reform.logging.exception.AlertLevel;
+import uk.gov.hmcts.reform.logging.exception.UnknownErrorCodeException;
+
+public class UnsupportedLetterRequestTypeException extends UnknownErrorCodeException {
+    public UnsupportedLetterRequestTypeException() {
+        super(AlertLevel.P1, "Unsupported lette request type");
```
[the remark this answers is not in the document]
rkondratowicz: yeah, if this is thrown something really really bad happened ;)

PR #99, opened 2018-04-06
Feature: AspectJ orientated dependency tracking

doncem:
Convert all dependency tracking into aspect point-cuts:

- completely get rid of the clutter
- magic

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
```diff
@@ -25,40 +33,37 @@ public AppInsights(TelemetryClient telemetry) {
...
-            AppDependencyCommand.FTP_REPORT_DELETE,
-            new Duration(duration.toMillis()),
-            success
-        );
+    @Around("dependencyPointCut(dependency)")
+    public Object trackDependency(ProceedingJoinPoint joinPoint, Dependency dependency) throws Throwable {
+        Instant start = Instant.now();
+
+        try {
+            Object proceed = joinPoint.proceed();
```
rkondratowicz: is this variable needed?

src/main/java/uk/gov/hmcts/reform/sendletter/services/AuthService.java
```diff
@@ -3,39 +3,25 @@
...
-
-                insights.trackServiceAuthentication(Duration.between(start, Instant.now()), true);
-
-                return serviceName;
-            } catch (Throwable exception) {
-                insights.trackServiceAuthentication(Duration.between(start, Instant.now()), false);
-
-                throw exception;
-            }
+            return authTokenValidator.getServiceName(authHeader);
```
rkondratowicz: ❤️

src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
doncem: yes. as can be expected in the java-logging too, think of a generalised version. this is a wrapper of any kind of method which is annotated with pointcut. method must return same thing whatever it returns

PR #102, opened 2018-04-09
Scheduled task interval cleanup

rkondratowicz:
- increase upload task delay in integration tests (every 1sec instead of every 1ms (sic!)
- make config value name clearer
- reduce wait time in e2e test (failing test would run for full 2min, now 30sec)
- fix code formatting

### DIFF
--- a/src/integrationTest/resources/application.yaml
+++ b/src/integrationTest/resources/application.yaml
@@ -41,7 +41,7 @@ ftp:
 scheduling.enabled: false
 
 tasks:
-  upload-letters: ${UPLOAD_LETTERS_INTERVAL:1}
+  upload-letters: ${UPLOAD_LETTERS_INTERVAL:1000} #
   # Run every second.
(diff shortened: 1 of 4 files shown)


src/integrationTest/resources/application.yaml
```diff
@@ -41,7 +41,7 @@ ftp:
 scheduling.enabled: false
 
 tasks:
-  upload-letters: ${UPLOAD_LETTERS_INTERVAL:1}
+  upload-letters-interval-ms: ${UPLOAD_LETTERS_INTERVAL:1000} #
```
nitinprabhu: Do we need `#` at the end?

nitinprabhu: Also we could update `UPLOAD_LETTERS_INTERVAL` to `UPLOAD_LETTERS_INTERVAL_IN_MS` ?

lgonczar: Would be nicer 👍

PR #103, opened 2018-04-09
Feature toggle scheduling with env var

rkondratowicz:


infrastructure/prod.tfvars
```diff
@@ -1,2 +1,3 @@
 vault_section = "prod"
 s2s_url = "https://prod-s2s-api.reform.hmcts.net:3511"
+scheduling_enabled = "false"
```
rkondratowicz: for now, we don't want 2 apps parsing the same reports on prod

PR #104, opened 2018-04-09
Move pdf files to collided package

doncem:

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/sendletter/services/PdfCreator.java
```diff
@@ -54,6 +55,7 @@ public PdfCreator(DuplexPreparator duplexPreparator, IHtmlToPdfConverter convert
         }
     }
 
+    // TODO dependency?
```
rkondratowicz: ?

doncem: should've been left alone locally

PR #106, opened 2018-04-10
Feature: PDF dependency

doncem:
Notes:

- Remove extra constructor from letter service which is not used anymore
- Split Pdf creator to separate beans because Spring AOP cannot properly recognise proxied class method call with built-in `@Dependency`

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/controllers/SendLetterTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/controllers/SendLetterTest.java
@@ -25,9 +25,12 @@ import uk.gov.hmcts.reform.sendletter.logging.AppInsights;
 import uk.gov.hmcts.reform.sendletter.logging.Dependency;
 
 import java.io.IOException;
+import java.util.List;
 
+import static java.util.stream.Collectors.toList;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoMoreInteractions;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppDependency.java
```diff
@@ -15,6 +15,11 @@
      */
     public static final String FTP_CLIENT = "FtpClient";
 
+    /**
+     * Pdf client.
+     */
+    public static final String PDF_CLIENT = "PdfClient";
```
rkondratowicz: we're not calling pdf service anymore

src/test/java/uk/gov/hmcts/reform/sendletter/services/PdfCreatorTest.java
```diff
@@ -37,7 +37,7 @@
 
     @Before
     public void setUp() {
-        pdfCreator = new PdfCreator(this.duplexPreparator, this.converter);
+        pdfCreator = new PdfCreator(this.duplexPreparator, new PdfConverter(this.converter));
```
rkondratowicz: mocking PdfConverter seems more straightforward

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/controllers/SendLetterTest.java
```diff
@@ -56,12 +59,20 @@ public void should_return_200_when_single_letter_is_sent() throws Throwable {
...
+        List<String> dependencyCommands = dependencies.stream().map(Dependency::command).collect(toList());
+
+        assertThat(dependencyNames).containsExactly(
+            AppDependency.AUTH_SERVICE,
+            AppDependency.PDF_CLIENT
+        );
+        assertThat(dependencyCommands).containsExactly(
+            AppDependencyCommand.AUTH_SERVICE_HEADER,
+            AppDependencyCommand.PDF_CREATE
+        );
```
rkondratowicz: can you just check that the `dependencies` list contains expected objects instead of extracting its' fields into separate lists?

doncem: I am not aware of annotation interface instantiation. If exists must be non trivial

rkondratowicz: you don't need to.

http://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#extracted-properties-assertion

```
assertThat(fellowshipOfTheRing).extracting("name", "age", "race.name")
                               .contains(
                                         tuple("Boromir", 37, "Man"),
                                         tuple("Sam", 38, "Hobbit"),
                                         tuple("Legolas", 1000, "Elf")
                               );
```

doncem: it's a different object used in the example. annotation interfaces are different concepts. here's error output for your example ^:

```bash
Can't find any field or property with name 'value'.
Error when introspecting properties was :
- No getter for property 'value' in com.sun.proxy.$Proxy181 
```

rkondratowicz: It crashed because there is no `getValue()` method in your class.
Try `.extractingResultOf(...)` instead (example in same link, scroll down).

PR #108, opened 2018-04-10
Fix temp file issue and filename

nitinprabhu:

https://tools.hmcts.net/jira/browse/RPE-228

### Change description ###

- Filename passed to temp file is the input zip file and not the output pgp file.

- `File.createTempFile` appends random long number to file which we don't want.Common's IO lib to the rescue.

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/reform/sendletter/encryption/PgpEncryptionUtil.java
```diff
@@ -35,8 +36,7 @@ private PgpEncryptionUtil() {
 
     public static byte[] encryptFile(
         byte[] inputFile,
-        String fileNamePrefix,
-        String fileNameSuffix,
+        String fileName,
```
rkondratowicz: is this the input file name or result file name?
Javadoc would be nice

src/test/java/uk/gov/hmcts/reform/sendletter/encryption/PgpEncryptionUtilTest.java
```diff
@@ -21,14 +21,10 @@ public void should_encrypt_and_create_pgp_encrypted_zip_file_when_valid_public_k
...
 
-        String fileNamePrefix = "encrypted";
-        String fileNameSuffix = "pgp";
-
         //when
         byte[] pgpEncryptedZip = PgpEncryptionUtil.encryptFile(
             inputZipFile,
-            fileNamePrefix,
-            fileNameSuffix,
+            "encrypted.zip",
```
rkondratowicz: file name is different that the name of the file that we pass. is this how this class is going to be used in real life?

nitinprabhu: you can use any file name in our real service we would be passing something similar to `CMC001_cmcclaimstore_22032018152746_d07c653b-9bcd-4c54-85ca-e22aa321d501.zip`

src/main/java/uk/gov/hmcts/reform/sendletter/encryption/PgpEncryptionUtil.java
```diff
@@ -34,9 +34,24 @@
     private PgpEncryptionUtil() {
     }
 
+    /**
+     * Encrypts the given byte array using PGP encryption using Triple DES algorithm.
+     * This method assumes that the temporary volume is writable
+     *
+     * @param inputFile          input file byte array
+     * @param inputFileName      input file name to be encrypted
+     * @param pgpPublicKey       public key to encrypt
```
rkondratowicz: encrypt with

src/main/java/uk/gov/hmcts/reform/sendletter/encryption/PgpEncryptionUtil.java
```diff
@@ -33,10 +34,24 @@
     private PgpEncryptionUtil() {
     }
 
+    /**
+     * Encrypts the given byte array using PGP encryption using Triple DES algorithm.
```
rkondratowicz: let's use AES if possible (separate PR obv)

PR #110, opened 2018-04-10
Feature: Internal dependency

doncem:
Record internal dependencies as events to AppInsights reflecting same properties as normal dependency have

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/controllers/SendLetterTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/controllers/SendLetterTest.java
@@ -22,7 +22,7 @@ import org.springframework.transaction.annotation.Transactional;
 import uk.gov.hmcts.reform.sendletter.logging.AppDependency;
 import uk.gov.hmcts.reform.sendletter.logging.AppDependencyCommand;
 import uk.gov.hmcts.reform.sendletter.logging.AppInsights;
-import uk.gov.hmcts.reform.sendletter.logging.Dependency;
+import uk.gov.hmcts.reform.sendletter.logging.ExternalDependency;
 
 import java.io.IOException;
 
@@ -49,16 +49,16 @@ public class SendLetterTest {
 
     @Test
     public void should_return_200_when_single_letter_is_sent() throws Throwable {
-        ArgumentCaptor<Dependency> dependencyCaptor = ArgumentCaptor.forClass(Dependency.class);
+        ArgumentCaptor<ExternalDependency> dependencyCaptor = ArgumentCaptor.forClass(ExternalDependency.class);
 
         MvcResult result = send(readResource("letter.json"))
             .andExpect(status().isOk())
             .andReturn();
 
         assertThat(result.getResponse().getContentAsString()).isNotNull();
-        verify(insights).trackDependency(any(ProceedingJoinPoint.class), dependencyCaptor.capture());
+        verify(insights).trackExternalDependency(any(ProceedingJoinPoint.class), dependencyCaptor.capture());
 
-        Dependency dependency = dependencyCaptor.getValue();
+        ExternalDependency dependency = dependencyCaptor.getValue();
 
         assertThat(dependency.value()).isEqualTo(AppDependency.AUTH_SERVICE);
         assertThat(dependency.command()).isEqualTo(AppDependencyCommand.AUTH_SERVICE_HEADER);
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
@@ -88,7 +88,7 @@ public class AppInsights extends AbstractAppInsights {
             telemetry.trackEvent(
                 dependency.value(),
                 ImmutableMap.of("success", "true"),
-                ImmutableMap.of("timeTook", ((double) MILLIS.between(start, Instant.now())) / 1000)
+                ImmutableMap.of("timeTaken", (double) MILLIS.between(start, Instant.now()))
             );
 
             return proceed;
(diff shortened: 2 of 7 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
```diff
@@ -66,6 +70,40 @@ public Object trackDependency(ProceedingJoinPoint joinPoint, Dependency dependen
...
+    ) throws Throwable {
+        Instant start = Instant.now();
+
+        try {
+            Object proceed = joinPoint.proceed();
+
+            telemetry.trackEvent(
+                dependency.value(),
+                ImmutableMap.of("success", "true"),
+                ImmutableMap.of("timeTook", ((double) MILLIS.between(start, Instant.now())) / 1000)
```
lgonczar: how about `timeTaken`?

src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
```diff
@@ -66,6 +70,40 @@ public Object trackDependency(ProceedingJoinPoint joinPoint, Dependency dependen
...
+        InternalDependency dependency
+    ) throws Throwable {
+        Instant start = Instant.now();
+
+        try {
+            Object proceed = joinPoint.proceed();
+
+            telemetry.trackEvent(
+                dependency.value(),
+                ImmutableMap.of("success", "true"),
```
rkondratowicz: can this be removed?

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/controllers/SendLetterTest.java
```diff
@@ -49,16 +49,16 @@
 
     @Test
     public void should_return_200_when_single_letter_is_sent() throws Throwable {
```
nitinprabhu: We are using Around Advice which means it gets invoked before and after the method call.It will good if that is evident in the test case.

src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
```diff
@@ -34,12 +35,15 @@ public AppInsights(TelemetryClient telemetry) {
     // dependencies
 
     @Pointcut("@annotation(dependency)")
```
nitinprabhu: Do we need to update this `ExternalDependency` or `InternalDependency`?

src/main/java/uk/gov/hmcts/reform/sendletter/logging/InternalDependency.java
```diff
@@ -0,0 +1,21 @@
...
+ * Annotation to be used against methods to be picked up by AppInsights aspect dependency advice.
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface InternalDependency {
+
+    /**
+     * Name of the internal dependency to be tracked as an event in Azure AppInsights.
+     *
+     * @return Name of the event
```
rkondratowicz: event or dependency?

doncem: The description formally states that **internal dependency** is tracked as an **event**, hence the event name

src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
doncem: No, the `dependency` is referring to argument name not the type. Unfortunately, cannot unify to use generic case

src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
```diff
@@ -66,6 +70,40 @@ public Object trackDependency(ProceedingJoinPoint joinPoint, Dependency dependen
...
+            );
+
+            return proceed;
+        } catch (Throwable exception) {
+            telemetry.trackEvent(
+                dependency.value(),
+                ImmutableMap.of("success", "false"),
+                ImmutableMap.of("timeTaken", (double) MILLIS.between(start, Instant.now()))
+            );
+            telemetry.trackException((Exception) exception);
```
rkondratowicz: not a part of this PR, but is this casting safe?

doncem: If `Error` is being caught here then essentially it is a bad thing and should be thrown. No application should experience `Error`

lgonczar: Actually, I think we should catch Exception instead of Throwable.

doncem: We should. `Throwable` comes from `jointPoint.proceed()`. Oh wait, just realised I am throwing it anyway and marking method as throwable. I'll update those advices

src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
```diff
@@ -34,12 +35,15 @@ public AppInsights(TelemetryClient telemetry) {
...
         // point cut definition
     }
 
-    @Around("dependencyPointCut(dependency)")
-    public Object trackDependency(ProceedingJoinPoint joinPoint, Dependency dependency) throws Throwable {
+    @Around("externalDependencyPointCut(dependency)")
+    public Object trackExternalDependency(
+        ProceedingJoinPoint joinPoint,
+        ExternalDependency dependency
+    ) throws Throwable {
```
rkondratowicz: this can now be `Exception` after your recent change I think

PR #112, opened 2018-04-10
Add Jenkinsfile for sandbox deployment

banderous:
### Change description ###
Add a parameterized Jenkinsfile based on producer repository.

```
[ ] Yes
[x] No
```


Jenkinsfile_parameterized
```diff
@@ -0,0 +1,34 @@
...
+    name: secretName,
+    version: '',
+    envVariable: envVar
+  ]
+}
+
+properties([
+  parameters([
+    string(name: 'PRODUCT_NAME', defaultValue: 'send-letter', description: ''),
+    string(name: 'APP', defaultValue: 'send-letter-service', description: ''),
```
lgonczar: I think this should just be `service`.  The pipeline appends it to the product name.

Jenkinsfile_parameterized
```diff
@@ -0,0 +1,34 @@
+#!groovy
+@Library("Infrastructure") _
+
+List<LinkedHashMap<String, Object>> secrets = [
+  secret('smoke-test-s2s-name', 'SMOKE_TEST_S2S_NAME'),
```
lgonczar: These keys are currently not set by terraform, which will cause the job to fail. You would need to specify them in main.tf, just like it's done in the producer.

banderous: I've updated the main terraform script.

infrastructure/main.tf
```diff
@@ -92,4 +97,16 @@ resource "azurerm_key_vault_secret" "smoke-test-s2s-url" {
   value     = "${var.s2s_url}"
   vault_uri = "${module.key-vault.key_vault_uri}"
 }
+
+resource "azurerm_key_vault_secret" "smoke-test-s2s-name" {
+  name      = "smoke-test-s2s-name"
+  value     = "send_letter_tests"
```
rkondratowicz: Is this service meant to be used in both functional and smoke tests?
If so, it would be worth renaming the secret.

PR #114, opened 2018-04-11
Add smoke test

rkondratowicz:
Creates letter, then check its' status.

Tested with locally running letter-service and s2s


src/smokeTest/resources/application-localdev.properties
```diff
@@ -0,0 +1,5 @@
+test-url=http://localhost:8485
+
+s2s-url=http://localhost:8489
+s2s-name=divorce
+s2s-secret=AAAAAAAAAAAAAAAA
```
rkondratowicz: for your convenience, if  you want to run it on local machine

PR #130, opened 2018-04-12
Clean up logging in upload job

rkondratowicz:
See comments on diff

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
@@ -8,7 +8,6 @@ import org.springframework.transaction.annotation.Transactional;
 import uk.gov.hmcts.reform.sendletter.entity.Letter;
 import uk.gov.hmcts.reform.sendletter.entity.LetterRepository;
 import uk.gov.hmcts.reform.sendletter.entity.LetterStatus;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -61,24 +60,17 @@ public void run() {
                 uploadLetter(letter);
             } catch (FtpException exception) {
                 logger.error(String.format("Exception uploading letter %s", letter.getId()), exception);
-
                 break;
-            } catch (DocumentZipException exception) {
-                logger.error(String.format("Failed to zip document for letter %s", letter.getId()), exception);
```
rkondratowicz: we no longer zip file here, this exception is never thrown

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -61,24 +60,17 @@ public void run() {
...
     private void uploadLetter(Letter letter) {
-        String uploadedFilename = uploadToFtp(letter);
+        uploadToFtp(letter);
 
-        logger.info(
-            "Successfully uploaded letter {}. File name: {}",
-            letter.getId(),
-            uploadedFilename
-        );
+        logger.info("Successfully uploaded letter {}", letter.getId());
```
rkondratowicz: we log the file name before it's sent, no need to repeat imho

PR #133, opened 2018-04-12
Use SSL on database connections

banderous:
### Change description ###
Create terraform db_connection_options variable and append to Postgres connection strings including ssl=true.

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -38,18 +38,18 @@ module "send-letter-service" {
   subscription        = "${var.subscription}"
 
   app_settings = {
-    S2S_URL                       = "${var.s2s_url}"
-    LETTER_TRACKING_DB_HOST       = "${module.db.host_name}"
-    LETTER_TRACKING_DB_PORT       = "${module.db.postgresql_listen_port}"
-    LETTER_TRACKING_DB_USER_NAME  = "${module.db.user_name}"
-    LETTER_TRACKING_DB_PASSWORD   = "${module.db.postgresql_password}"
-    LETTER_TRACKING_DB_NAME       = "${module.db.postgresql_database}"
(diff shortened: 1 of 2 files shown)


src/main/resources/application.yaml
```diff
@@ -13,7 +13,7 @@ spring:
   application:
     name: Send Letter Provider
   datasource:
-      url: jdbc:postgresql://${LETTER_TRACKING_DB_HOST:send-letter-producer-database}:${LETTER_TRACKING_DB_PORT:5432}/${LETTER_TRACKING_DB_NAME:letter_tracking}
+      url: jdbc:postgresql://${LETTER_TRACKING_DB_HOST:send-letter-producer-database}:${LETTER_TRACKING_DB_PORT:5432}/${LETTER_TRACKING_DB_NAME:letter_tracking}${LETTER_TRACKING_DB_CONN_OPTIONS}
```
doncem: can we enable default here as an empty value?

infrastructure/main.tf
```diff
@@ -40,7 +44,8 @@ module "send-letter-service" {
     LETTER_TRACKING_DB_USER_NAME  = "${module.db.user_name}"
     LETTER_TRACKING_DB_PASSWORD   = "${module.db.postgresql_password}"
     LETTER_TRACKING_DB_NAME       = "${module.db.postgresql_database}"
-    FLYWAY_URL                    = "jdbc:postgresql://${module.db.host_name}:${module.db.postgresql_listen_port}/${module.db.postgresql_database}"
+    LETTER_TRACKING_DB_CONN_OPTIONS = "${local.db_connection_options}"
```
doncem: Don't want to be that guy, but I saw in different PRs people like the vertical alignment in terraform files

PR #146, opened 2018-04-16
Include letter status in json response

rkondratowicz:
on GET requests


src/main/java/uk/gov/hmcts/reform/sendletter/model/out/LetterStatus.java
```diff
@@ -26,13 +28,15 @@
 
     public LetterStatus(
         final UUID id,
+        final String status,
         final String messageId,
         final ZonedDateTime createdAt,
         final ZonedDateTime sentToPrintAt,
         final ZonedDateTime printedAt,
         final boolean hasFailed
```
doncem: can we have a todo here - i think `hasFailed` is redundant now and can be removed in next pr

src/main/java/uk/gov/hmcts/reform/sendletter/model/out/LetterStatus.java
```diff
@@ -26,13 +28,15 @@
 
     public LetterStatus(
         final UUID id,
+        final String status,
```
doncem: shouldn't we be using enum here too?

src/main/java/uk/gov/hmcts/reform/sendletter/model/out/LetterStatus.java
banderous: Though if it is part of our public API that would be a breaking change.

PR #147, opened 2018-04-16
Added functional tests and terraform config

nitinprabhu:

https://tools.hmcts.net/jira/browse/RPE-379

### Change description ###

- Added functional test cases and terraform config to set environment variables to be used by functional tests.

- Tested functional test by pointing it to sandbox environment.

- Changes only cover the existing version where it accepts template and json data.

- Functional tests would be run only on prod pipeline so we need to bring that CNP Jenkins file  back and update it according to the parameterised file changes done for this PR.

```
[ ] Yes
[ x] No
```


Jenkinsfile_parameterized
```diff
@@ -5,6 +5,15 @@ List<LinkedHashMap<String, Object>> secrets = [
   secret('test-s2s-name', 'TEST_S2S_NAME'),
   secret('test-s2s-secret', 'TEST_S2S_SECRET'),
   secret('test-s2s-url', 'TEST_S2S_URL')
+  secret('test-send-letter-url', 'TEST_SEND_LETTER_URL'),
```
lgonczar: This one is not needed. Jenkins will provide the value as `TEST_URL`.

infrastructure/main.tf
```diff
@@ -36,10 +36,13 @@ data "vault_generic_secret" "encryption_public_key" {
 }
 
 locals {
+  ase_name               = "${data.terraform_remote_state.core_apps_compute.ase_name[0]}"
   ftp_private_key        = "${replace(data.vault_generic_secret.ftp_private_key.data["value"], "\\n", "\n")}"
   ftp_public_key         = "${replace(data.vault_generic_secret.ftp_public_key.data["value"], "\\n", "\n")}"
   ftp_user               = "${data.vault_generic_secret.ftp_user.data["value"]}"
   encryption_public_key  = "${replace(data.vault_generic_secret.encryption_public_key.data["value"], "\\n", "\n")}"
+  s2s_url                = "http://rpe-service-auth-provider-${var.env}.service.${local.ase_name}.internal"
+  send_letter_url        = "http://send-letter-service-${var.env}.service.${local.ase_name}.internal"
```
lgonczar: We can probably get rid of this, as well

PR #148, opened 2018-04-16
Feature: track events and metrics to AppInsights

doncem:
Notes:

- event for mark letters posted task
- metric for upload letters task
- minor cleanup of unchecked exceptions in test cases which are no longer requried
- super minor log event change

Throughout codebase log events are sufficient enough events for app insights. Agent configurations covers essential key points of the applications. Introduced two metric values figuratively 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
@@ -46,7 +46,7 @@ public class AppInsights extends AbstractAppInsights {
         telemetry.trackEvent(LETTER_NOT_PRINTED, properties, null);
     }
 
-    public void trackPrintReport(ParsedReport report) {
+    public void trackPrintReportReceived(ParsedReport report) {
         telemetry.trackEvent(
             LETTER_PRINT_REPORT,
             ImmutableMap.of("isReportParsedFully", BooleanUtils.toStringYesNo(report.allRowsParsed)),
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
@@ -1,9 +1,12 @@
 package uk.gov.hmcts.reform.sendletter.logging;
 
+import com.google.common.collect.ImmutableMap;
 import com.microsoft.applicationinsights.TelemetryClient;
+import com.microsoft.applicationinsights.core.dependencies.apachecommons.lang3.BooleanUtils;
 import org.springframework.stereotype.Component;
 import uk.gov.hmcts.reform.logging.appinsights.AbstractAppInsights;
 import uk.gov.hmcts.reform.sendletter.entity.Letter;
+import uk.gov.hmcts.reform.sendletter.model.ParsedReport;
 
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
@@ -15,6 +18,8 @@ import java.util.Map;
 public class AppInsights extends AbstractAppInsights {
 
     static final String LETTER_NOT_PRINTED = "LetterNotPrinted";
+
+    static final String LETTER_PRINT_REPORT = "LetterPrintReport";
 
     static final DateTimeFormatter TIME_FORMAT = DateTimeFormatter.ofPattern("HH:mm:ss");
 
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
```diff
@@ -47,6 +51,7 @@ public void run(LocalTime now) {
                 .stream()
                 .map(parser::parse)
                 .forEach(parsedReport -> {
+                    insights.trackPrintReport(parsedReport);
```
rkondratowicz: it would be safer to log this after everything is done (few lines below) I think

src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
```diff
@@ -16,6 +19,10 @@
 
     static final String LETTER_NOT_PRINTED = "LetterNotPrinted";
 
+    static final String LETTER_PRINT_REPORT = "LetterPrintReport";
```
banderous: Could we use an enum for these values?

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -52,17 +56,24 @@ public void run() {
...
                 try {
                     uploadLetter(letter);
+                    counter++;
                 } catch (FtpException exception) {
                     logger.error(String.format("Exception uploading letter %s", letter.getId()), exception);
                     break;
                 }
             }
 
+            if (counter > 0) {
```
banderous: Could we not also report zero letters uploaded so we can see the task is running?

banderous: I don't mean as a separate event, just removing the if.

src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
doncem: Enum is not really valuable in this case. We still need string as a telemetry name and there is nothing to compare to or keep any control of it functionally.

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
doncem: isn't log event sufficient enough to assume task is running? there are several logs happening during:
```java
// scheduler
String name = task.name();
log.info("Trying to lock {}", name);
log.info("Acquired lock {}", name);
// task
logger.info("Started letter upload job");
logger.info("Completed letter upload job");
// back to scheduler
log.info("Released lock {}", name);
```

src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
banderous: It would be nice if App insights took an object as an event name and called toString on it, then one could pass either a string or an enum to it. Nevermind though.

banderous: I'm not sure LetterPrintReport makes sense as an event name because it doesn't have a verb; perhaps LetterPrintReportProcessed or LetterReportProcessed or XeroxReportProcessed?

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
banderous: It depends what this event is tracking; is it tracking that we successfully processed a report or that we started processing a report?

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
banderous: I'm wondering how it looks in app insights but can't seem to see any events in there to guide me.

My understanding is that logging is not available in app insights, so if you saw that this event was not occurring you would not know if it was because there is no work to do or because there is a problem with the task. Accessing logs is a pain, so it would be nice if App insights gave us a good overview of system health.

It may be nothing needs to be done here since i don't know what app insights looks like yet so I defer to you.

PR #153, opened 2018-04-19
Use pagination in upload letters task

banderous:
### Change description ###
Hibernate's streams must be called from within a transaction, which is undesirable for batch processing tasks since the transaction will be rolled back if there is a problem anywhere in the job, including for entities that were processed successfully.

This change processes letters in batches of 10.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
@@ -129,7 +129,7 @@ public class UploadLettersTaskTest {
 
     @Test
     public void should_process_all_letter_batches() throws Exception {
-        IntStream.rangeClosed(1, 20).forEach(
+        IntStream.rangeClosed(1, UploadLettersTask.BATCH_SIZE * 2).forEach(
             x -> letterService.send(SampleData.letterRequest(), "service"));
 
         // and
--- a/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
@@ -129,7 +129,8 @@ public class UploadLettersTaskTest {
 
     @Test
     public void should_process_all_letter_batches() throws Exception {
-        IntStream.rangeClosed(1, UploadLettersTask.BATCH_SIZE * 2).forEach(
+        int letterCount = UploadLettersTask.BATCH_SIZE * 2;
+        IntStream.rangeClosed(1, letterCount).forEach(
             x -> letterService.send(SampleData.letterRequest(), "service"));
 
         // and
@@ -143,6 +144,6 @@ public class UploadLettersTaskTest {
             task.run();
         }
         entityManager.clear();
-        assertThat(repository.findByStatus(LetterStatus.Uploaded).count()).isEqualTo(20);
+        assertThat(repository.findByStatus(LetterStatus.Uploaded).count()).isEqualTo(letterCount);
     }
 }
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -41,7 +41,6 @@ public UploadLettersTask(
         this.availabilityChecker = availabilityChecker;
     }
 
-    @Transactional
```
nitinprabhu: Wouldn't this be easier if we had moved this to `uploadLetter` and make that method public?

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
```diff
@@ -123,4 +126,23 @@ public void should_fail_to_upload_to_sftp_and_stop_from_uploading_any_other_lett
...
+            repository,
+            FtpHelper.getSuccessfulClient(LocalSftpServer.port),
+            availabilityChecker
+        );
+
+        try (LocalSftpServer server = LocalSftpServer.create()) {
+            task.run();
+        }
+        entityManager.clear();
+        assertThat(repository.findByStatus(LetterStatus.Uploaded).count()).isEqualTo(20);
```
rkondratowicz: is 20 the same as BATCH_SIZE * 2?

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
```diff
@@ -123,4 +126,23 @@ public void should_fail_to_upload_to_sftp_and_stop_from_uploading_any_other_lett
...
+        UploadLettersTask task = new UploadLettersTask(
+            repository,
+            FtpHelper.getSuccessfulClient(LocalSftpServer.port),
+            availabilityChecker
+        );
+
+        try (LocalSftpServer server = LocalSftpServer.create()) {
+            task.run();
+        }
+        entityManager.clear();
```
rkondratowicz: may be worth adding `// Clear the JPA cache to force a read.` here as well. (or perhaps extract to a method?)

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
banderous: If all the letters are marked Uploaded in the same transaction then they either all be committed or all rolled back; having a second transaction won't change this.

Each letter needs to be a separate transaction which is why streams don't work, with streams the order of events is as follows:

BEGIN TRANSACTION
for each letter in stream {
  upload letter
  UPDATE LETTER to Uploaded
}
COMMIT
Nothing will be committed until the iteration completes and the outer transaction commits, and if anything goes wrong the transaction is rolled back.

I'm fairly sure the only way around this is to break this up into multiple transactions. See the [last comment here](https://jira.spring.io/browse/DATAJPA-1023)

banderous: Though we might be talking about different things; I still need to fix the cause of the SQLException in sandbox, that requires a separate investigation.

banderous: Still, the streams approach is not going to work if we have 100,000 letters to upload.

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTaskTest.java
banderous: good spot

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
```diff
@@ -9,6 +11,8 @@
 
 public interface LetterRepository extends JpaRepository<Letter, UUID> {
 
+    Page<Letter> findByStatus(LetterStatus status, Pageable pageable);
+
     Stream<Letter> findByStatus(LetterStatus status);
```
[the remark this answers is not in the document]
banderous: Yes but only by tests. I tried getting rid of it but it makes for quite awkward test code.

PR #154, opened 2018-04-20
Improve flyway Dockerfile

doncem:
tiny update avoids the step of downloading postgres client

```
[ ] Yes
[x] No
```

### DIFF
--- a/docker/flyway/Dockerfile
+++ b/docker/flyway/Dockerfile
@@ -2,10 +2,13 @@ FROM postgres:9.6
 
 WORKDIR /flyway
 
-ADD http://repo1.maven.org/maven2/org/flywaydb/flyway-commandline/5.0.7/flyway-commandline-5.0.7.zip /flyway.zip
+ENV FLYWAY_VERSION 5.0.7
 
-RUN unzip /flyway.zip && rm /flyway.zip && mv /flyway-5.0.7 /flyway && ln -s /flyway/flyway /usr/local/bin/flyway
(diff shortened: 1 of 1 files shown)


docker/flyway/Dockerfile
```diff
@@ -1,13 +1,15 @@
-FROM openjdk:8-jre
+FROM postgres:9.6
 
-RUN apt-get update && apt-get -y install postgresql-client
+WORKDIR /flyway
 
-ADD http://repo1.maven.org/maven2/org/flywaydb/flyway-commandline/5.0.7/flyway-commandline-5.0.7.zip /flyway.zip
+ENV FLYWAY_VERSION 5.0.7
```
nitinprabhu: Can we move this to `.env` please ?

docker/flyway/Dockerfile
```diff
@@ -1,13 +1,15 @@
...
 
-RUN unzip /flyway.zip && rm /flyway.zip && mv /flyway-5.0.7 /flyway && ln -s /flyway/flyway /usr/local/bin/flyway
+ADD http://repo1.maven.org/maven2/org/flywaydb/flyway-commandline/${FLYWAY_VERSION}/flyway-commandline-${FLYWAY_VERSION}.tar.gz flyway-commandline-${FLYWAY_VERSION}.tar.gz
```
nitinprabhu: would be great if we could use flyway image just that if the url is not available for some reason the image building breaks.Not that worried as it is just for local.

docker/flyway/Dockerfile
doncem: Dockerfile does not recognise `.env` ;)

PR #156, opened 2018-04-20
Fix e2e integration tests

doncem:
Fails to load app context due to missing config

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/resources/application.yaml
+++ b/src/integrationTest/resources/application.yaml
@@ -48,4 +48,4 @@ tasks:
 
 encryption:
   enabled: ${ENCRYPTION_ENABLED:false}
-  publicKey: ${ENCRYPTION_PUBLIC_KEY}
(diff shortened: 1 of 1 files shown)


src/integrationTest/resources/application.yaml
```diff
@@ -48,4 +48,4 @@ tasks:
 
 encryption:
   enabled: ${ENCRYPTION_ENABLED:false}
-  publicKey: ${ENCRYPTION_PUBLIC_KEY}
+  publicKey: ${ENCRYPTION_PUBLIC_KEY:}
```
nitinprabhu: e2e test are run as part of integration tests.This env var should be set imo https://github.com/hmcts/send-letter-service/blob/master/build.gradle#L74? So do we need to do this?

doncem: something happened and it stopped working really: master/prs/local integration run via gradle task is not working

PR #163, opened 2018-04-25
Add local cache for prod

doncem:

[Investigate solution to MS 502 errors](https://tools.hmcts.net/jira/browse/RPE-389)

### Change description ###

[reference](https://github.com/hmcts/moj-module-webapp/pull/89)

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/prod.tfvars
+++ b/infrastructure/prod.tfvars
@@ -1,4 +1,4 @@
 vault_section = "prod"
 s2s_url = "https://prod-s2s-api.reform.hmcts.net:3511"
 scheduling_enabled = "false"
-local_cache = "Always"
+local_cache = "Off"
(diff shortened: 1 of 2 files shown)


banderous:
Can we enable on sandbox too?

infrastructure/prod.tfvars
```diff
@@ -1,4 +1,4 @@
 vault_section = "prod"
 s2s_url = "https://prod-s2s-api.reform.hmcts.net:3511"
 scheduling_enabled = "false"
-local_cache = "Always"
+local_cache = "Off"
```
nitinprabhu: I would advise make if default off and then turn if on per environment through <env>.tfvars file.A good candidate for sandbox to see if there is any difference in 502 errors.Most importantly are we seeing any 502's on this service as CMC has not started using this so how we are going to verify the fix?

PR #164, opened 2018-04-25
Infra config prod pipeline

nitinprabhu:
### Change description ###

- Added `Jenkins_CNP` file to build on prod pipeline.

- By default encryption is turned off and scheduling is turned off through `prod.tfvars`.

- Needs to be tested on preview environment which will done once PR is created.

```
[ ] Yes
[x ] No
```

### DIFF
--- a/infrastructure/variables.tf
+++ b/infrastructure/variables.tf
@@ -47,7 +47,7 @@ variable encyption_enabled {
 }
 
 variable scheduling_enabled {
-  default = "true"
+  default = "false"
 }
 
 variable "ftp_hostname" {
@@ -76,4 +76,8 @@ variable "ftp_reports_folder" {
 
 variable "ftp_reports_cron" {
   default = "0 30 0 * * *"
+}
+
+variable "deployment_namespace" {
+  default = ""
 }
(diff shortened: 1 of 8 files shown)


rkondratowicz:
it may be worth changing the default value in `application.yaml` to `false`, just in case...

infrastructure/variables.tf
```diff
@@ -77,3 +77,7 @@ variable "ftp_reports_folder" {
 variable "ftp_reports_cron" {
   default = "0 30 0 * * *"
 }
+
+variable "deployment_namespace" {
+  default = ""
+}
```
rkondratowicz: should it not be `variable "deployment_namespace" { }` instead?

nitinprabhu: I have followed what was done in rhubarb-service so not sure if we don't have default it would work.Can give it a try though.

nitinprabhu:
Merging this now as successfully deployed on preview.

src/integrationTest/resources/application.yaml
```diff
@@ -43,8 +43,8 @@ scheduling.enabled: false
 tasks:
   upload-letters-interval-ms: ${UPLOAD_LETTERS_INTERVAL:1000}
   # Run every second.
-  mark-letters-posted: ${FTP_REPORTS_CRON:*/1 * * * * *}
-  stale-letters-report: ${FTP_REPORTS_CRON:*/1 * * * * *}
+  mark-letters-posted: ${FTP_REPORTS_CRON:*/5 * * * * *}
+  stale-letters-report: ${FTP_REPORTS_CRON:*/5 * * * * *}
```
rkondratowicz: ?

PR #165, opened 2018-04-25
Bug - delay background jobs in e2e tests

rkondratowicz:
In some cases (quite often on travis) the jobs in e2e test would start before in-memory sftp server was up, thus causing exception which made the tests fail.

This PR introduces `FakeFtpAvailabilityChecker` class (used only in tests) which can be used to control the FTP 'downtime window' and prevent jobs from trying to make a connection to in-memory sftp server.

### DIFF
--- a/src/integrationTest/resources/application.yaml
+++ b/src/integrationTest/resources/application.yaml
@@ -43,8 +43,8 @@ scheduling.enabled: false
 tasks:
   upload-letters-interval-ms: ${UPLOAD_LETTERS_INTERVAL:1000}
   # Run every second.
-  mark-letters-posted: ${FTP_REPORTS_CRON:*/5 * * * * *}
-  stale-letters-report: ${FTP_REPORTS_CRON:*/5 * * * * *}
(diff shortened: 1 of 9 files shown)


src/integrationTest/resources/application.yaml
```diff
@@ -43,8 +43,8 @@ scheduling.enabled: false
 tasks:
   upload-letters-interval-ms: ${UPLOAD_LETTERS_INTERVAL:1000}
   # Run every second.
-  mark-letters-posted: ${FTP_REPORTS_CRON:*/5 * * * * *}
-  stale-letters-report: ${FTP_REPORTS_CRON:*/5 * * * * *}
+  mark-letters-posted: ${FTP_REPORTS_CRON:*/1 * * * * *}
+  stale-letters-report: ${FTP_REPORTS_CRON:*/1 * * * * *}
```
rkondratowicz: bringing back original value, slowing down jobs 'launch' was a temporary fix

PR #168, opened 2018-04-27
Updated type for zip/pgp file name for functional tests

nitinprabhu:
### Change description ###

- File name pattern generation logic has changed but functional tests were not updated.This PR updates that.

- PDF file name generation is unchanged so no changes for that.

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/sendletter/FunctionalTestSuite.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/sendletter/FunctionalTestSuite.java
@@ -31,8 +31,6 @@ import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
 @TestPropertySource("classpath:application.properties")
 public abstract class FunctionalTestSuite {
 
-    protected static final String SMOKE_TEST_LETTER_TYPE = "smoke_test";
-
     @Value("${s2s-url}")
     protected String s2sUrl;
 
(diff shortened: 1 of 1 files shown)


src/functionalTest/java/uk/gov/hmcts/reform/sendletter/FunctionalTestSuite.java
```diff
@@ -161,7 +159,7 @@ protected ZipInputStream getZipInputStream(RemoteFile zipFile) throws IOExceptio
     protected String getPdfFileNamePattern(String letterId) {
         return String.format(
             "%s_%s_%s.pdf",
-            Pattern.quote(SMOKE_TEST_LETTER_TYPE),
+            Pattern.quote("smoke_test"),
```
doncem: now as i look, is pattern correct? injecting a string with underscore here

src/functionalTest/java/uk/gov/hmcts/reform/sendletter/FunctionalTestSuite.java
```diff
@@ -172,7 +170,7 @@ protected String getFileNamePattern(String letterId) {
 
         return String.format(
             format,
-            Pattern.quote(SMOKE_TEST_LETTER_TYPE),
+            Pattern.quote("smoketest"),
```
rkondratowicz: did the test pass before???

nitinprabhu: I had tested it by posting the functional tests to sandbox but the service was updated after that.This is first time we tested on prod pipeline.If the file name generation logic was unchanged then we it would pass :)

PR #170, opened 2018-04-27
Added vault name and vault uri

nitinprabhu:
### Change description ###

- As we have not specified the vault name in our module default naming convention is <product>-<env>.

- As we have s2s in prod and aat and that one use default vault name we get a conflict 409  as `rpe-aat` and `rpe-prod`  already exists in those environment.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/infrastructure/output.tf
+++ b/infrastructure/output.tf
@@ -1,9 +1,9 @@
 output "vaultUri" {
-  value = "${module.key-vault.key_vault_uri}"
+  value = "${local.vaultUri}"
 }
 
 output "vaultName" {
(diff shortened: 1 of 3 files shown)


infrastructure/output.tf
```diff
@@ -1,9 +1,9 @@
 output "vaultUri" {
-  value = "${module.key-vault.key_vault_uri}"
+  value = "${local.vaultUri}"
```
lgonczar: I guess you could just refer straight to vault's output vault_uri param and get rid of this local one.

PR #173, opened 2018-04-30
Fix flakeyness in integration tests

banderous:
### Change description ###

I found two other causes of test flakeyness. The first was a race condition between our tests and the scheduled tasks, where the test would try and read and validate PDF files before they had been fully written to disk by the FTP server.

A second issue came from the MarkLettersPosted task, which attempts to connect to Xerox's FTP server, and during our tests sometimes tries and fails to do so either before our test FTP server
has started or after it has been shut down.

This change catches and logs FtpExceptions that occur during the task, instead of letting them propagate & fail the test since FtpExceptions are expected during normal operation and don't indicate a bug in the task.

With these changes I ran the integration tests around 50 times consecutively without a failure.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/e2e/BaseTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/e2e/BaseTest.java
@@ -74,7 +74,8 @@ public class BaseTest {
 
             // Wait for letter to be uploaded.
             await().atMost(15, SECONDS).untilAsserted(
-                () -> assertThat(validLettersUploaded(server.lettersFolder, isEncryptionEnabled)).as("No letters uploaded!").isTrue()
+                () -> assertThat(validLettersUploaded(server.lettersFolder, isEncryptionEnabled))
+                      .as("No letters uploaded!").isTrue()
             );
 
             // Generate Xerox report.
@@ -138,7 +139,7 @@ public class BaseTest {
                 );
             }
             PdfHelper.validateZippedPdf(content);
-        } catch (IOException|PGPException e) {
+        } catch (IOException | PGPException e) {
             // File may still be being written to disk by FTP server.
             return false;
         }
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
```diff
@@ -48,9 +49,12 @@ public MarkLettersPostedTask(
     }
 
     public void run(LocalTime now) {
-        if (ftpAvailabilityChecker.isFtpAvailable(now)) {
-            logger.info("Started '{}' task", MarkLettersPosted);
+        if (!ftpAvailabilityChecker.isFtpAvailable(now)) {
+            logger.info("Not processing '{}' task due to FTP downtime window", MarkLettersPosted);
+        }
```
doncem: Should actually not process instead of continuing to `try`

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/e2e/BaseTest.java
```diff
@@ -105,32 +106,43 @@ private boolean letterHasBeenPosted() {
         return letters.size() == 1 && letters.get(0).getStatus() == LetterStatus.Posted;
     }
 
-    private void createXeroxReport(LocalSftpServer server, Boolean isEncryptionEnabled) throws IOException {
-        try (Stream<UUID> letterIds = Arrays.stream(server.lettersFolder.listFiles())
-            .map(file -> validateGettingLetterId(file, isEncryptionEnabled))) {
+    private void createXeroxReport(LocalSftpServer server) throws IOException {
+        try (Stream<UUID> letterIds = Arrays.stream(server.lettersFolder.list())
+            .map(FileNameHelper::extractIdFromPdfName)) {
```
rkondratowicz: we upload .zip (or .pgp), not .pdf, and file name format for pdf and zip are different. what am I missing?

banderous: In both cases the letter id is the penultimate component of the name

PR #174, opened 2018-04-30
Move PostgreSQL database from West Europe to UK South

lgonczar:
### Change description ###

Move PostgreSQL database from West Europe to UK South. Also, because the new version of moj-module-postgres allows for it, rename the DB (to something different than `postgres`) and user.

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -54,7 +54,7 @@ locals {
 
 module "db" {
   source              = "git@github.com:hmcts/moj-module-postgres?ref=cnp-449-tactical"
-  product             = "${var.product}-postgres-db"
+  product             = "${var.product}-${var.component}-db"
   location            = "${var.location_db}"
   env                 = "${var.env}"
   database_name       = "send_letter"
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -54,11 +54,11 @@ locals {
 
 module "db" {
   source              = "git@github.com:hmcts/moj-module-postgres?ref=cnp-449-tactical"
-  product             = "${var.product}-db"
+  product             = "${var.product}-postgres-db"
   location            = "${var.location_db}"
   env                 = "${var.env}"
-  postgresql_database = "postgres"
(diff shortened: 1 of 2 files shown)


infrastructure/main.tf
```diff
@@ -53,12 +53,14 @@ locals {
 }
 
 module "db" {
-  source              = "git@github.com:hmcts/moj-module-postgres.git?ref=master"
+  source              = "git@github.com:hmcts/moj-module-postgres?ref=cnp-449-tactical"
```
rkondratowicz: can this be merged to master?

lgonczar: I suppose this has to be a controlled change. For now, we're supposed to use this one. My guess is that this branch will be merged into master once all DBs have been migrated.

infrastructure/main.tf
```diff
@@ -53,12 +53,14 @@ locals {
...
   product             = "${var.product}-db"
   location            = "${var.location_db}"
   env                 = "${var.env}"
-  postgresql_database = "postgres"
-  postgresql_user     = "letter_tracking"
+  database_name       = "send_letter"
+  postgresql_user     = "send_letter"
+  sku_name            = "GP_Gen5_2"
```
nitinprabhu: Does the sku name needs to be this ? or can we have a proper name?

infrastructure/main.tf
nitinprabhu: .git is missing here `git@github.com:hmcts/moj-module-postgres.git?ref=cnp-449-tactical`?

infrastructure/main.tf
lgonczar: I think this is the actual name for "General purpose" sku

infrastructure/main.tf
lgonczar: I also noticed it in the guideline, but decided to copy-paste it and see if the pipeline would break in preview. As the preview pipelines are all breaking now, I can't see that, yet.

PR #175, opened 2018-04-30
Simplify pgp util exceptions

rkondratowicz:
Throw runtime exceptions defined in the encryption module from the util class instead of forcing the clients to handle ambiguous `IOException` and bouncycastle's exception.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/services/encryption/PgpEncryptionUtil.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/services/encryption/PgpEncryptionUtil.java
@@ -14,6 +14,8 @@ import org.bouncycastle.openpgp.PGPUtil;
 import org.bouncycastle.openpgp.bc.BcPGPPublicKeyRing;
 import org.bouncycastle.openpgp.operator.bc.BcPGPDataEncryptorBuilder;
 import org.bouncycastle.openpgp.operator.bc.BcPublicKeyKeyEncryptionMethodGenerator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
@@ -30,6 +32,8 @@ import static org.bouncycastle.openpgp.PGPUtil.getDecoderStream;
 
 public final class PgpEncryptionUtil {
 
+    private static final Logger log = LoggerFactory.getLogger(PgpEncryptionUtil.class);
+
     // Prevent instantiation.
     private PgpEncryptionUtil() {
     }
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/services/encryption/PgpEncryptionUtil.java
```diff
@@ -44,43 +48,49 @@ private PgpEncryptionUtil() {
...
-            );
-
-        PGPEncryptedDataGenerator encryptedDataGenerator =
-            prepareDataEncryptor(
-                pgpPublicKey,
-                withIntegrityCheck
-            );
-
-        return writeEncryptedDataToOutputStream(byteArrayOutputStream, encryptedDataGenerator);
+    ) {
```
rkondratowicz: best viewed by adding `?w=1` to url

src/main/java/uk/gov/hmcts/reform/sendletter/services/encryption/PgpEncryptionUtil.java
```diff
@@ -44,43 +48,49 @@ private PgpEncryptionUtil() {
...
-                getDecoderStream(new ByteArrayInputStream(data))
-            )
-        ).orElseThrow(() -> new UnableToLoadPgpPublicKeyException("PGP Public key object could not be constructed"));
+    public static PGPPublicKey loadPublicKey(byte[] data) {
+        try {
+            return lookupPublicSubkey(
+                new BcPGPPublicKeyRing(
+                    getDecoderStream(new ByteArrayInputStream(data))
+                )
+            ).orElseThrow(() -> new UnableToLoadPgpPublicKeyException(null));
```
banderous: Unrelated to this change, but should UnableToLoadPgpPublicKeyException exist as an Exception? It seems more like an assertion failure that should immediately crash the app, it isn't an exceptional condition that can be handled.

src/main/java/uk/gov/hmcts/reform/sendletter/services/encryption/PgpEncryptionUtil.java
```diff
@@ -44,43 +48,49 @@ private PgpEncryptionUtil() {
...
+            PGPEncryptedDataGenerator encryptedDataGenerator =
+                prepareDataEncryptor(
+                    pgpPublicKey,
+                    withIntegrityCheck
+                );
+
+            return writeEncryptedDataToOutputStream(byteArrayOutputStream, encryptedDataGenerator);
+        } catch (IOException | PGPException exc) {
+            log.error("Error encrypting file", exc);
+            throw new UnableToPgpEncryptZipFileException(exc);
```
doncem: I would go for more abstract name, e.g. `UnableToEncryptFileException`. There is no need to be extremely verbose with exception names

src/main/java/uk/gov/hmcts/reform/sendletter/services/encryption/PgpEncryptionUtil.java
doncem: Same here, could do with just `UnableToLoadPublicKeyException`. Not even pgp related then. specifically for any attempt to load public key

PR #176, opened 2018-05-02
Feature: Spring Boot 2

doncem:

[Investigate migration to Spring Boot 2](https://tools.hmcts.net/jira/browse/RPE-394)

### Change description ###

~Depends on [http-proxy-spring-boot](https://github.com/hmcts/http-proxy-spring-boot-autoconfigure/pull/3) repo.~

The rest hmcts artifacts are in spring boot 2 already - bumping.

Changes in application configuration are followed by migration guide (see link below).

~Dependency suppressions will be re-checked once the migration is accomplished.~

~HELP! with e2e test. Perhaps @banderous ?~

[Further reading](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.0-Migration-Guide) on migration guide

```
[x] Yes
[ ] No
```


src/main/resources/application.yaml
```diff
@@ -1,10 +1,6 @@
 server:
   port: 8485
 
-management:
-  security:
-    enabled: ${MANAGEMENT_SECURITY_ENABLED:true}
```
rkondratowicz: is there are different config for this now?

doncem: `management.security.*` has been completely removed and only can be editable via java code. since we not even using the spring boot security package - i don't think this config was even applied at any point in the past

banderous:
LGTM 👍

build.gradle
```diff
@@ -243,15 +240,19 @@ dependencies {
...
+bootJar {
+  archiveName = 'send-letter-service.jar'
 
   manifest {
     attributes('Implementation-Version': project.version.toString())
   }
 }
 
-task migratePostgresDatabase(type: org.flywaydb.gradle.task.FlywayMigrateTask) {
+import org.flywaydb.gradle.task.FlywayMigrateTask
```
nitinprabhu: do we need to do this ?

doncem: as part of upgrade: solve soft warnings in gradle scripts

doncem:
good point. but to answer the question directly, no - i don't want :). will add that config

PR #185, opened 2018-05-04
Fix sonar smells

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/SerialTaskRunner.java
```diff
@@ -97,8 +95,9 @@ private boolean executeReturningBool(Connection connection, String sql) throws S
         try (ResultSet set = connection.createStatement().executeQuery(sql)) {
             if (set.next()) {
                 return set.getBoolean(1);
+            } else {
+                return false;
```
doncem: now this block looks a lot like simple return statement: `return set.next();`

rkondratowicz: not really, we `.getBoolean()` may still be have `false` value

doncem: oh yea, where am i looking¿

PR #187, opened 2018-05-08
Fix void method reporting

doncem:
Application Insights' Agent did not know how to track void methods. Need to specifically specify signature as mentioned in example [here](https://github.com/MicrosoftDocs/azure-docs/blob/master/articles/application-insights/app-insights-java-agent.md#configure-the-agent) and how-to [guide](http://www.rgagnon.com/javadetails/java-0286.html)

```
[ ] Yes
[x] No
```

### DIFF
--- a/lib/AI-Agent.xml
+++ b/lib/AI-Agent.xml
@@ -19,18 +19,18 @@
             <Method name="encryptFile"/>
         </Class>
         <Class name="uk.gov.hmcts.reform.sendletter.services.ftp.FtpClient">
-            <Method name="upload"/>
+            <Method name="upload" signature="(Lnet/schmizz/sshj/xfer/LocalSourceFile;Z)V"/>
             <Method name="downloadReports"/>
-            <Method name="deleteReport"/>
+            <Method name="deleteReport" signature="(Ljava/lang/String)V"/>
         </Class>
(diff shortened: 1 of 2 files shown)


lib/AI-Agent.xml
```diff
@@ -19,18 +19,18 @@
             <Method name="encryptFile"/>
         </Class>
         <Class name="uk.gov.hmcts.reform.sendletter.services.ftp.FtpClient">
-            <Method name="upload"/>
+            <Method name="upload" reportExecutionTime="true" signature="(Lnet/schmizz/sshj/xfer/LocalSourceFile;Z)V"/>
             <Method name="downloadReports"/>
-            <Method name="deleteReport"/>
+            <Method name="deleteReport" reportExecutionTime="true" signature="(Ljava/lang/String;)V"/>
```
rkondratowicz: what's the actual fix?
adding `reportExecutionTime="true"` or specifying signature?
isn't signature only needed when there are two methods with the same name?

doncem:
Even though it is the correct "by the book" integration, but still seems to not work. We need that reporting in place so will do the old way - manually request dependency tracking as we did previously in the consumer.

Will draft separate PR

PR #188, opened 2018-05-09
Add info endpoint

rkondratowicz:
shows number of letters in given status


src/main/java/uk/gov/hmcts/reform/sendletter/info/LetterStatusInfoContributor.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Override
+    public void contribute(Info.Builder builder) {
+        builder.withDetail(
+            "letters_by_status",
+            Stream.of(LetterStatus.values())
+                .collect(
+                    toMap(
+                        s -> s.name().toLowerCase(),
+                        s -> repo.countByStatus(s)
```
rkondratowicz: I've started with single query (`group by...`) but it required a custom native query and a new entity model, so I went with this istead.

PR #189, opened 2018-05-09
Fix FTP dependency tracking

doncem:
As inspected in production - there is no FTP tracking happening automagically for us. After attempts to apply signature as MS suggests no luck achieved. Going back to old way of just manually tracking the dependency to keep track of this activity

```
[ ] Yes
[x] No
```

### DIFF
--- a/lib/AI-Agent.xml
+++ b/lib/AI-Agent.xml
@@ -18,11 +18,6 @@
         <Class name="uk.gov.hmcts.reform.sendletter.services.encryption.PgpEncryptionUtil">
             <Method name="encryptFile"/>
         </Class>
-        <Class name="uk.gov.hmcts.reform.sendletter.services.ftp.FtpClient">
-            <Method name="upload"/>
-            <Method name="downloadReports"/>
-            <Method name="deleteReport"/>
-        </Class>
         <Class name="uk.gov.hmcts.reform.sendletter.services.pdf.PdfCreator">
             <Method name="createFromTemplates"/>
             <Method name="createFromBase64Pdfs"/>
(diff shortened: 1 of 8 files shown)


lib/AI-Agent.xml
```diff
@@ -18,11 +18,6 @@
         <Class name="uk.gov.hmcts.reform.sendletter.services.encryption.PgpEncryptionUtil">
```
banderous: Are any of these being tracked automatically or can we get rid of this file?

src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/FtpClient.java
```diff
@@ -80,19 +95,31 @@ public void upload(LocalSourceFile file, boolean isSmokeTestFile) {
                         }
                     })
                     .collect(toList());
+
+                insights.trackFtpReportDownload(Duration.between(now, Instant.now()), true);
```
banderous: you could combine these last two insights tracking calls by using a finally block after the exception handler, here and in a couple of other places

lib/AI-Agent.xml
doncem: Encryption and pdf creator tracks - no problem. for some reason a mixture of void methods are not working. Next would be task scheduler + silencing info logs. Only need to know single hit of the task and it's success rate

src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/FtpClient.java
```diff
@@ -52,9 +63,13 @@ public void upload(LocalSourceFile file, boolean isSmokeTestFile) {
                 String path = String.join("/", folder, file.getName());
                 sftp.getFileTransfer().upload(file, path);
 
+                success = true;
```
banderous: I hadn't spotted that we tracked success! Happy with either approach.

doncem: yea, only two key features of dependency: success and time took

src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/FtpClient.java
```diff
@@ -32,18 +35,26 @@
...
         this.configProperties = configProperties;
+        this.insights = insights;
     }
     // endregion
 
     public void upload(LocalSourceFile file, boolean isSmokeTestFile) {
+        Instant now = Instant.now();
+
         runWith(sftp -> {
+            boolean success = false;
```
nitinprabhu: can we rename this to isSuccess please ?

PR #195, opened 2018-06-04
Exclude smoke tests from stale letter report

doncem:
### Change description ###

Smoke tests piled up as stale letters gives false information about what's actually stale. Excluding from query

stale-letters

P.S. Spike is from data migration

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
@@ -1,6 +1,8 @@
 package uk.gov.hmcts.reform.sendletter.entity;
 
 import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.query.Param;
 
 import java.sql.Timestamp;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
```diff
@@ -14,7 +16,12 @@
 
     Stream<Letter> findByStatus(LetterStatus status);
 
-    Stream<Letter> findByStatusAndSentToPrintAtBefore(LetterStatus status, Timestamp before);
+    @Query("select l from Letter l where l.status = :status and l.type <> :type and l.sentToPrintAt < :before")
+    Stream<Letter> findStaleLetters(
+        @Param("status") LetterStatus status,
+        @Param("type") String type,
```
banderous: Do we need to parameterise the type here? I would just hard code the smoke test type here.

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
```diff
@@ -14,7 +17,13 @@
 
     Stream<Letter> findByStatus(LetterStatus status);
 
-    Stream<Letter> findByStatusAndSentToPrintAtBefore(LetterStatus status, Timestamp before);
+    @Query("select l from Letter l where l.status = :status and l.type <> '"
+        + UploadLettersTask.SMOKE_TEST_LETTER_TYPE
+        + "' and l.sentToPrintAt < :before")
+    Stream<Letter> findStaleLetters(
+        @Param("status") LetterStatus status,
```
rkondratowicz: either keep the old method name, or remove the `status` parameter.

Perhaps rename and remove param on a future PR?

PR #202, opened 2018-09-05
[PREVIEW] Update gradle build

doncem:
### Change description ###

Before introducing any changes introducing separate PR with non-major version jumps (java-logging includes major app insights version jump which is non breaking but just not including as part of this PR). Follow up:

- plugin major jump is not considered as breaking
- gradle update left aside together with app insights
- test dependencies should always be up-to-date and not considered as breaking change to the app itself
- hystrix super major jump. that's just aligning to actual spring boot v2 which is already there
- fix in s2s client with v1.0.1 where it does not have spring cloud milestone in place

```
[ ] Yes
[x] No
```


docker-compose-flyway.yml
```diff
@@ -6,7 +6,7 @@ services:
       file: docker-compose-core.yml
       service: service
     environment:
-      - FLYWAY_ENABLED=true
+      - SPRING_FLYWAY_ENABLED=true
```
nitinprabhu: oh this changed in boot 2 👍

PR #229, opened 2018-11-01
[PREVIEW] Update alert query

doncem:
#### Notes

- trying to automate pr build by including some dummy email for alert which is required by terraform plan/apply step
- improve alert query

#### Summary

Query overrides requirement of time window in which the alerting condition is checked. Module requires specific window in minutes and only judge by that. Incorporating it in the query by 5 additional windows making the interval equal to 3 hours instead.
Same measurements are taken and every half an hour and summarising to anticipated `is_operational` flag.
Update gives opportunity to increase threshold which can be interpreted as _correct_ for given failure/alert: only stream of them makes valid assumption whether one can question service is down or not

```
[ ] Yes
[x] No
```

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -31,11 +31,14 @@ withPipeline(type , product, component) {
 
   enableDbMigration()
   enableSlackNotifications(channel)
-  loadVaultSecrets(secrets)
 
   onPR {
-    loadVaultSecrets([
+    loadVaultSecrets(secrets + [
       secret('send-letter-failure-email', 'TEST_FAILURE_EMAIL')
     ])
+  }
+
+  onNonPR {
+    loadVaultSecrets(secrets)
   }
 }
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -32,4 +32,10 @@ withPipeline(type , product, component) {
   enableDbMigration()
   enableSlackNotifications(channel)
   loadVaultSecrets(secrets)
+
+  onPR {
+    loadVaultSecrets([
+      secret('send-letter-failure-email', 'TEST_FAILURE_EMAIL')
+    ])
+  }
 }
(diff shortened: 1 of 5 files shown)


infrastructure/output.tf
```diff
@@ -40,4 +40,8 @@ output "test_encryption_enabled" {
   value = "${var.encyption_enabled}"
 }
 
+output "test_failure_email" {
+  value = "ignore@failure.email"
```
timja: What's this for?

doncem: to be used in preview prs. so it goes _smoothly_. temporary value anyway - no need for alerts actually show up in preview

timja: but its an output? where is it used?

doncem: i thought jenkinsfile uses these values when `loadVaultSecrets`

timja: its outputted for tests, its only there for smoke and functional tests

Jenkinsfile_CNP
```diff
@@ -31,5 +31,14 @@ withPipeline(type , product, component) {
 
   enableDbMigration()
   enableSlackNotifications(channel)
-  loadVaultSecrets(secrets)
+
+  onPR {
+    loadVaultSecrets(secrets + [
+      secret('send-letter-failure-email', 'TEST_FAILURE_EMAIL')
```
lgonczar: How is `TEST_FAILURE_EMAIL` used

doncem: it's a through away value used only for PRs so [this action group](https://github.com/hmcts/send-letter-service/blob/master/infrastructure/is_down_alert_action_group.tf#L15) can be created. automatically from first run

lgonczar: I can see that the checks have passed, but I don't quite understand why. I thought `secret` only creates an environment variable (in this case `TEST_FAILURE_EMAIL`) based on the Azure Key Vault secret (`send-letter-failure-email`) which must exist. The action group definition references the secret.

doncem: based on previous comment from @timja my understanding also has been revoked. now i think it just sets the string in case actual env var is not found ¯\\\_(ツ)_/¯

PR #234, opened 2018-11-16
Fix OWASP issue around Guava dependencies

doncem:
### Change description ###

- Use latest java-logging library
- Use stable non-shady guava library
- Suppress shady guava 20.0 library coming together with app insights

```
[ ] Yes
[x] No
```


timja:
I vaguely remember a problem with deleting files in use, around the app insights agent jar.
You may need to stage this PR to upgrade it

i.e. add new one in this PR
delete the old one in the following PR

doncem:
Really? Don't remember this being an issue before. I do though remember Microsoft pointing out a problem if multiple agent jars are kept in same directory 🤔

timja:
Give it a go, it you get an access denied problem you'll know why 👅

PR #243, opened 2018-12-04
Feature: validate request with base64 pdfs

doncem:

[Validate base64 encoded pdf for send letter service endpoint accepting pdf documents](https://tools.hmcts.net/jira/browse/BPS-256)

### Change description ###

Making sure that invalid request is marked as `400 Bad Request`.

This PR introduces the feature from jackson which automatically decodes base64 strings into byte arrays therefore automatically validating.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/sendletter/services/pdf/PdfCreator.java
```diff
@@ -34,15 +32,13 @@ public PdfCreator(DuplexPreparator duplexPreparator, IHtmlToPdfConverter convert
         return PdfMerger.mergeDocuments(docs);
     }
 
-    public byte[] createFromBase64Pdfs(List<String> base64encodedDocs) {
-        Asserts.notNull(base64encodedDocs, "base64encodedDocs");
+    public byte[] createFromBase64Pdfs(List<byte[]> base64decodedDocs) {
+        Asserts.notNull(base64decodedDocs, "base64decodedDocs");
```
nitinprabhu: is any other test case covering this ? If not could we please add it

doncem: yes. all the same unit/integration tests plus additional integration tests introduced when base64 is received but not a pdf. this method is basically simplifying the approach since decoding happens upon deserialisation of the incoming request rather than further down the line as before (which was here in this file).

PR #245, opened 2018-12-04
[PREVIEW] Extract template to separate html file

doncem:
### Change description ###

Preparing functional resources to make future PR simpler to read

```
[ ] Yes
[x] No
```


doncem:
For some reason app returns 400 without any clue why and just couldn't dig the problem why. Will close and leave for future, perhaps will be better approach enlightened later on

doncem:
Just one more try with `[PREVIEW]`. I have a fear `abstract` test suite causes the break for #246

PR #266, opened 2018-12-19
Continue processing other files if folder for given file is not configured

rkondratowicz:
just in case...

Removing the logic of determining the target folder from FTP client.
The caller sends the FTP client service folder that should be used.
When folder for given service is not found, this fact is logged and the client is never called.

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/services/FtpUploadTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/services/FtpUploadTest.java
@@ -24,7 +24,7 @@ public class FtpUploadTest {
         FileToSend doc = new FileToSend("hello.zip", "world".getBytes());
         try (LocalSftpServer server = LocalSftpServer.create()) {
             FtpClient client = FtpHelper.getSuccessfulClient(LocalSftpServer.port);
-            client.upload(doc, false, "bulkprint");
+            client.upload(doc, false, LocalSftpServer.SERVICE_FOLDER);
             File[] files = server.lettersFolder.listFiles();
             assertThat(files.length).isEqualTo(1);
             String content = new String(Files.toByteArray(files[0]));
--- a/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/services/FtpUploadTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/services/FtpUploadTest.java
@@ -2,7 +2,6 @@ package uk.gov.hmcts.reform.sendletter.services;
 
 import com.google.common.io.Files;
 import org.junit.Test;
-import uk.gov.hmcts.reform.sendletter.exception.ServiceNotConfiguredException;
 import uk.gov.hmcts.reform.sendletter.helper.FtpHelper;
 import uk.gov.hmcts.reform.sendletter.services.ftp.FileToSend;
 import uk.gov.hmcts.reform.sendletter.services.ftp.FtpClient;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/FtpClient.java
```diff
@@ -51,15 +49,13 @@ public FtpClient(
...
 
-    public void upload(LocalSourceFile file, boolean isSmokeTestFile, String service) {
+    public void upload(LocalSourceFile file, boolean isSmokeTestFile, String serviceFolder) {
         Instant now = Instant.now();
 
         runWith(sftp -> {
             boolean isSuccess = false;
 
             try {
-                String serviceFolder = getServiceFolderMapping(service);
```
lgonczar: 👍

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -79,20 +84,25 @@ public void run() {
...
+            );
+
+            ftp.upload(file, isSmokeTest(letter), serviceFolder.get());
+
+            logger.info(
+                "Uploaded letter id: {}, messageId: {}, file name: {}, additional data: {}",
+                letter.getId(),
+                letter.getMessageId(),
+                file.filename,
+                letter.getAdditionalData()
```
lgonczar: Is additional data not sensitive information?

nitinprabhu: Those should claim reference or ccd case reference and it helps for debugging and analysing issues on prod. Services should not send any sensitive data afaik :)

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -79,20 +84,25 @@ public void run() {
...
-
-        logger.info(
-            "Uploaded letter id: {}, messageId: {}, file name: {}, additional data: {}",
-            letter.getId(),
-            letter.getMessageId(),
-            file.filename,
-            letter.getAdditionalData()
-        );
+        Optional<String> serviceFolder = serviceFolderMapping.getFolderFor(letter.getService());
+        if (serviceFolder.isPresent()) {
```
lgonczar: `ifPresent` could be used instead. Not much difference, up to you.

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/services/FtpUploadTest.java
```diff
@@ -26,28 +24,11 @@ public void uploads_file() throws Exception {
...
-            client.upload(doc, false, "bulkprint");
+            client.upload(doc, false, LocalSftpServer.SERVICE_FOLDER);
             File[] files = server.lettersFolder.listFiles();
             assertThat(files.length).isEqualTo(1);
             String content = new String(Files.toByteArray(files[0]));
             assertThat(content).isEqualTo("world");
         }
     }
-
-    @Test
```
lgonczar: We could also rename the name of this test, as it seems to be the client test. Just a suggestion.

rkondratowicz: you mean the class?

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
rkondratowicz: it can't, I need an `else`. `ifPresent` returns void.

lgonczar: true, what was I thinking?

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/services/FtpUploadTest.java
lgonczar: sorry, yes, test class

PR #274, opened 2019-01-04
Feature: AKS builds

doncem:

[Implement AKS for send letter service](https://tools.hmcts.net/jira/browse/BPS-399)

### Change description ###

mostly trying to follow bulk scan processor as it has db set up there and we **only** have db in this service. which we control. there's ftp factor to cover

```
[ ] Yes
[x] No
```


kubernetes/deployment.template.yaml
```diff
@@ -0,0 +1,178 @@
...
+---
+#
+# Send Letter Service Config
+#
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: send-letter-config
+  namespace: ${NAMESPACE}
+data:
```
timja: if you use used helm instead this config would be 1 tenth the size...
This is deprecated new projects shouldn't use it

See this draft store PR for the implementation:
https://github.com/hmcts/draft-store/pull/243/files

doncem: thanks for the link. please follow my explanation in jira ticket. in case of a strong position in using helm charts - please move back to `in progress` with a comment follow-up

timja: Have done so

doncem: thank you

PR #297, opened 2019-01-28
Rename message_id to checksum: Step 4

rkondratowicz:
### Start reading from the new column

1. Add new column to DB ✅ https://github.com/hmcts/send-letter-service/pull/290
2. Start writing to both old and new column, keep reading from old column ✅ https://github.com/hmcts/send-letter-service/pull/294
3. Copy data from old column to new column ✅ https://github.com/hmcts/send-letter-service/pull/295
4. **Start reading from the new column**
5. Stop writing to the old column
6. Delete old column
7. Update REST API (add extra field)

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
@@ -25,7 +25,7 @@ public interface LetterRepository extends JpaRepository<Letter, UUID> {
         @Param("before") Timestamp before
     );
 
-    Optional<Letter> findByMessageIdAndStatusOrderByCreatedAtDesc(String messageId, LetterStatus status);
+    Optional<Letter> findByChecksumAndStatusOrderByCreatedAtDesc(String checksum, LetterStatus status);
 
     Optional<Letter> findById(UUID id);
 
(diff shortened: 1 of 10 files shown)


src/main/resources/db/migration/V013__Make_message_id_nullable.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE letters
+ALTER COLUMN message_id DROP NOT NULL;
```
doncem: dropping should be separate pr as reading (application) is published after the migration is done

rkondratowicz: this is dropping a constraint, column becomes nullable, it's fine.

doncem: ignore me :|

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
```diff
@@ -25,7 +25,7 @@
         @Param("before") Timestamp before
     );
 
-    Optional<Letter> findByMessageIdAndStatusOrderByCreatedAtDesc(String messageId, LetterStatus status);
+    Optional<Letter> findByChecksumAndStatusOrderByCreatedAtDesc(String checksum, LetterStatus status);
```
lgonczar: This change means that currently deployed version searches for letters by message ID. This PR starts populating checksums only, which means that the old version will not be able to find letters created by the new version. This change in the repository should be done first.

rkondratowicz: done

PR #304, opened 2019-02-01
Update flyway

doncem:
### Change description ###

Removing test for flyway migration capabilities - it is capable, there's no need to test that. Introducing unit test which checks versions are correct so we can be extra sure next merge will be ok.

p.s. jenkins already merges master but nevertheless

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/reform/sendletter/data/migration/DbMigrationFilesTest.java
```diff
@@ -0,0 +1,45 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashSet;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class DbMigrationFilesTest {
+
+    @Test
+    public void should_have_unique_migrations() throws IOException {
```
rkondratowicz: cool idea 👍

doncem:
> extra points if it also checks consecutive numbers are used ;)

ha!

PR #317, opened 2019-02-20
Update alert with BSP shared infra and latest module changes

doncem:

Part of [create alerts for application errors](https://tools.hmcts.net/jira/browse/BPS-550)

### Change description ###

Group email is created. Applying change to send letter service so the action group can be updated with new email address instead of platform team

```
[ ] Yes
[x] No
```


doncem:
Actually, we have that now but it is defined in bulk scan shared infra

timja:
ah ok, do you still need the action group here then?

doncem:
Nope. I think I've created this PR before started working on shared infra. Adding WIP and will try to stick that in

PR #322, opened 2019-03-01
Update java logging

doncem:
### Change description ###

Latest update is transition to spring boot. It also has shaded guava vulnerability fix.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
```diff
@@ -24,27 +23,28 @@
...
     }
 
     public void trackFtpReportDeletion(java.time.Duration duration, boolean success) {
-        trackDependency(AppDependency.FTP_CLIENT, AppDependencyCommand.FTP_REPORT_DELETED, duration, success);
+        trackDependency(AppDependencyCommand.FTP_REPORT_DELETED, duration, success);
     }
 
     public void trackFtpReportDownload(java.time.Duration duration, boolean success) {
-        trackDependency(AppDependency.FTP_CLIENT, AppDependencyCommand.FTP_REPORT_DOWNLOADED, duration, success);
+        trackDependency(AppDependencyCommand.FTP_REPORT_DOWNLOADED, duration, success);
```
rkondratowicz: are those changes related to library upgrade?

doncem: Please forgive for not outlining this in the description. As part of the upgrade one should notice there's a small boy scout task available in the same file. Absolutely harmless one might add.

If it is still up to a challenge for a new PR - by all means I will revert this change

PR #334, opened 2019-03-15
Add service for deleting old files

rkondratowicz:
Not scheduled/used yet.

https://tools.hmcts.net/jira/browse/BPS-578

PR #335, opened 2019-03-18
Clean up old files on AAT periodically

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-578

Run the service created in https://github.com/hmcts/send-letter-service/pull/334 every day at 5am to delete files that are older than 12 hours (on AAT only)

### DIFF
--- a/infrastructure/prod.tfvars
+++ b/infrastructure/prod.tfvars
@@ -1,5 +1,6 @@
 vault_section = "prod"
 scheduling_enabled = "true"
 encyption_enabled = "true"
+file_cleanup_enabled = "false"
 
 capacity = "2"
--- a/infrastructure/variables.tf
+++ b/infrastructure/variables.tf
@@ -69,6 +69,10 @@ variable scheduling_enabled {
   default = "false"
 }
 
+variable "file_cleanup_enabled" {
+  default = "false"
+}
+
 variable scheduling_lock_at_most_for {
   default = "PT10M"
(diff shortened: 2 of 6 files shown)


infrastructure/prod.tfvars
```diff
@@ -1,5 +1,6 @@
 vault_section = "prod"
 scheduling_enabled = "true"
 encyption_enabled = "true"
+file_cleanup_enabled = "false"
```
doncem: no need - default will be used

infrastructure/variables.tf
```diff
@@ -69,6 +69,10 @@ variable scheduling_enabled {
   default = "false"
 }
 
+variable "file_cleanup_enabled" {
+  default = "false"
```
doncem: would be nice to have short description :)

infrastructure/prod.tfvars
rkondratowicz: just in case default changes 😱

infrastructure/variables.tf
rkondratowicz: done

infrastructure/prod.tfvars
doncem: tru dat. i wonder whether we could stick in some smoke test checking the cleanup component is loaded in context or not 🤔. but then it will mean it's running on prod staging slot

PR #336, opened 2019-03-18
Update README

doncem:
### Change description ###

Reproduced steps for dockerised environment. Need to bring back portion of deleted files from #284.

Bonus:

- 9cc962f
- 929b86a

```
[ ] Yes
[x] No
```


README.md
```diff
@@ -5,91 +5,58 @@
...
+$ docker run -d -i -t --name bsp-send-letter-database --hostname send-letter-database --env-file ./docker/env.list send-letter-database
 ```
 
-This will start the API container exposing the application's port
-(set to `8485` in this app).
+Important note: `--hostname send-letter-database` will be used to link it to service `run` command and also used with default value provided in application config.
+There is no need to provide full `--env-file` for the DB container. Only `LETTER_TRACKING_DB_PASSWORD=password` is required
 
-In order to test if the application is up, you can call its health endpoint:
+Launch the service from image hosted in Azure:
```
nitinprabhu: Azure -> Azure Container Registry :)

docker/database/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM postgres:9.6
```
nitinprabhu: can use alpine image ?

PR #338, opened 2019-03-20
Update the rest of dependencies

doncem:
### Change description ###

- gradle bump 4.10.2 -> 5.2.1. soon 5.3 will be available
- shedlock bump 2.3 -> 2.4. fixed deprecated config presumingly candidate for removal in next major jump. also moving `@EnableScheduling` to configuration instead where condition on property exists therefore enable will happen only if flag is on rather than per application scope

Note. Not sure why e2e tests locally did not work. One thing noticed is invalid properties for public/private keys used. Still did not work so will hope it's just my machine and pipeline will go green 👍

```
[ ] Yes
[x] No
```


gradlew
```diff
@@ -28,7 +28,7 @@ APP_NAME="Gradle"
 APP_BASE_NAME=`basename "$0"`
 
 # Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS=""
+DEFAULT_JVM_OPTS='"-Xmx64m"'
```
nitinprabhu: Is there a reason we are setting this now ?

gradlew.bat
```diff
@@ -14,7 +14,7 @@ set APP_BASE_NAME=%~n0
 set APP_HOME=%DIRNAME%
 
 @rem Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-set DEFAULT_JVM_OPTS=
+set DEFAULT_JVM_OPTS="-Xmx64m"
```
nitinprabhu: Is there a reason we are setting this now ?

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/e2e/BaseTest.java
```diff
@@ -150,4 +154,12 @@ private boolean isValidPdf(boolean isEncryptionEnabled, File file) {
         }
         return true;
     }
+
+    @EnableConfigurationProperties(FtpConfigProperties.class)
+    static class TestConfiguration {
```
nitinprabhu: Not sure why we need this ?

doncem: in config values are empty and library spat out "empty file" error when i tried to run these tests. I can try removing this but not sure how it works in the pipeline

gradlew
doncem: `gradlew` and `gradlew.bat` automatically generated after `$ gradle wrapper` command run ;). so i would expect this is OK

src/main/java/uk/gov/hmcts/reform/sendletter/config/SchedulerConfiguration.java
```diff
@@ -2,40 +2,26 @@
...
+import org.springframework.scheduling.annotation.EnableScheduling;
 
-import java.time.Duration;
 import javax.sql.DataSource;
 
 @Configuration
 @AutoConfigureAfter(FlywayConfiguration.class)
 @DependsOn({"flyway", "flywayInitializer"})
 @ConditionalOnProperty(value = "scheduling.enabled", matchIfMissing = true)
+@EnableScheduling
```
nitinprabhu: Do you need this ? It's already here https://github.com/hmcts/send-letter-service/blob/master/src/main/java/uk/gov/hmcts/reform/sendletter/Application.java#L10

doncem: moved it to relevant place. comment on description:
> also moving `@EnableScheduling` to configuration instead where condition on property exists therefore enable will happen only if flag is on rather than per application scope

PR #353, opened 2019-03-28
Enable SSCS

doncem:
### Change description ###

Onboard SSCS team

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -63,6 +63,8 @@ ftp:
       folder: DIVORCE
     - service: probate_backend
       folder: PROBATE
+    - service: sscs
+      folder: SSCS
 
 scheduling:
   enabled: ${SCHEDULING_ENABLED:false}


src/main/resources/application.yaml
```diff
@@ -63,6 +63,8 @@ ftp:
       folder: DIVORCE
     - service: probate_backend
       folder: PROBATE
+    - service: sscs
```
nitinprabhu: please check the s2s service name

ryandac: @jackmaloney do you know what an s2s service name is?

doncem: logs suggested this was the service accessing send letter

ryandac: I'm reliably informed it's set here https://github.com/hmcts/sscs-evidence-share/blob/master/src/main/resources/application.yaml#L54 and is usually sscs

PR #355, opened 2019-04-05
Adjust cron for jobs related to csv files

rkondratowicz:
CSV files are uploaded to FTP at 10am.


doncem:
please rebase on demo and merge at your own will ;). have just switched default branch for the time being

PR #360, opened 2019-04-11
Adjust ftp downtime window

rkondratowicz:


src/main/resources/application.yaml
```diff
@@ -52,7 +52,7 @@ ftp:
   privateKey: ${FTP_PRIVATE_KEY}
   publicKey: ${FTP_PUBLIC_KEY}
   downtime:
-    from: ${FTP_DOWNTIME_FROM:16:00}
+    from: ${FTP_DOWNTIME_FROM:15:00}
     to: ${FTP_DOWNTIME_TO:17:00}
```
doncem: missed second time?

doncem: @nitinprabhu pointed out it doesn't really matter as long as that specific hour is skipped 👍

PR #363, opened 2019-04-17
Use latest (9.0.19) of embed tomcat

doncem:
### Change description ###

New vulnerability discovered including `9.0.17` of tomcat embed coming in along with spring boot framework. Jumping _bugfix_ version to latest fixes the issue

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -198,6 +198,12 @@ dependencyManagement {
     dependencySet(group: 'org.mockito', version: versions.mockitoJupiter) {
       entry 'mockito-core'
     }
+    // CVE-2019-0232 - command line injections on windows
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -198,6 +198,12 @@ dependencyManagement {
     dependencySet(group: 'org.mockito', version: versions.mockitoJupiter) {
       entry 'mockito-core'
     }
+    // CVE-2019-0232 - command line injections on windows
+    dependencySet(group: 'org.apache.tomcat.embed', version: '9.0.19') {
```
timja: maybe consider bumping the tomcat version used on the ASE too?
https://github.com/hmcts/cnp-module-webapp/blob/master/variables.tf#L154
it defaults to latest 8.x

PR #364, opened 2019-04-17
Specify tomcat major version for webapp

doncem:
### Change description ###

Webapp module defaults to `8.0`. Bumping config to use the one from spring boot. See @timja comment #363

```
[ ] Yes
[x] No
```


doncem:
merging as there's no code change - no need for static tests nor aks builds

PR #366, opened 2019-04-23
Update test dependencies

doncem:
### Change description ###

- latest pitest (been dragging this one for quite some time now)
- jupiter updates:
  - junit 5.4.1 -> 5.4.2
  - mockito 2.25.1 -> 2.27.0

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -182,8 +182,8 @@ repositories {
 def versions = [
   bouncycastle: '1.61',
   flywayV: plugins.getPlugin('org.flywaydb.flyway').class.package.implementationVersion,
-  junit: '5.4.1',
-  mockitoJupiter: '2.25.1',
+  junit: '5.4.2',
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -141,6 +141,10 @@ sonarqube {
   }
 }
 
+pitest {
+  pitestVersion = '1.4.7'
```
rkondratowicz: line 12 has `id 'info.solidsoft.pitest' version '1.4.0'`

doncem: plugin is latest: `1.4.0`. the tool version differs and from default `1.4.3` pulled in by plugin and dependency management we now specify latest which is `1.4.7`. similarly like pmd plugin except this one has two misleading version notions :|

nitinprabhu: Do we need pitest plugin and it's dependencies ?

PR #368, opened 2019-04-24
Use `Europe/London` timezone in tasks

doncem:
### Change description ###

Depending what date it is `Europe/London` fluctuates between **GMT** and **BST**. Attaching zone ids to cron tasks and to ftp availability checks

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
@@ -17,7 +17,10 @@ import uk.gov.hmcts.reform.sendletter.services.ftp.FtpClient;
 import uk.gov.hmcts.reform.sendletter.services.ftp.IFtpAvailabilityChecker;
 
 import java.time.LocalTime;
+import java.time.ZoneId;
 import java.util.Optional;
+
+import static uk.gov.hmcts.reform.sendletter.util.TimeZones.EUROPE_LONDON;
 
 /**
  * Fetches reports from SFTP concerning posted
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
```diff
@@ -51,9 +54,9 @@ public MarkLettersPostedTask(
     }
 
     @SchedulerLock(name = TASK_NAME)
-    @Scheduled(cron = "${tasks.mark-letters-posted}")
+    @Scheduled(cron = "${tasks.mark-letters-posted}", zone = EUROPE_LONDON)
     public void run() {
-        if (!ftpAvailabilityChecker.isFtpAvailable(LocalTime.now())) {
+        if (!ftpAvailabilityChecker.isFtpAvailable(LocalTime.now(ZoneId.of(EUROPE_LONDON)))) {
```
rkondratowicz: Maybe this logic of using London time should live inside the `ftpAvailabilityChecker`?

doncem: can be. was considering about this

doncem:
`BST` is during `British Summer Time` only @lgonczar. After October it goes back to `GMT` 😂

rkondratowicz:
you can pull this story in https://tools.hmcts.net/jira/browse/BPS-605

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
doncem: Quickly stabbed and inspected it is not as trivial as passing local time with zone id as it is done here. Each time used in checker then has to be converted to `OffsetTime` instead

rkondratowicz: at least it could be tested

PR #369, opened 2019-04-24
Update configuration for RestTemplate

doncem:
### Change description ###

Following [suggestion](https://howtodoinjava.com/spring-restful/resttemplate-httpclient-java-config/) and the fact that app insights tracks http calls made only via apache http, this configuration should provide such functionality

Couple bonus commits just to drag in some super minor amendments

```
[ ] Yes
[x] No
```


doncem:
Fun fact. With these changes integration tests finally pass on my local. I don't remember last time experiencing such event

PR #370, opened 2019-04-24
Improving try/catch block accompanying exception message

doncem:
### Change description ###

`SSHClient` is of `Closeable` interface so can be entirely moved to _try with resources_. Also fixing exception message. Apologies for not splitting this into separate commit

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/FtpClient.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/FtpClient.java
@@ -4,8 +4,6 @@ import net.schmizz.sshj.SSHClient;
 import net.schmizz.sshj.sftp.RemoteResourceInfo;
 import net.schmizz.sshj.sftp.SFTPClient;
 import net.schmizz.sshj.sftp.SFTPFileTransfer;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.stereotype.Component;
 import uk.gov.hmcts.reform.sendletter.config.FtpConfigProperties;
@@ -17,6 +15,7 @@ import uk.gov.hmcts.reform.sendletter.model.Report;
 import java.io.IOException;
 import java.time.Duration;
 import java.time.Instant;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
 import java.util.function.Function;
@@ -27,8 +26,6 @@ import static java.util.stream.Collectors.toList;
 @Component
 @EnableConfigurationProperties(FtpConfigProperties.class)
 public class FtpClient {
-
-    private static final Logger logger = LoggerFactory.getLogger(FtpClient.class);
 
     private final FtpConfigProperties configProperties;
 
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/FtpClient.java
```diff
@@ -187,15 +180,10 @@ public void testConnection() {
...
             }
         } catch (IOException exc) {
-            throw new FtpException("Unable to upload file.", exc);
-        } finally {
-            try {
-                if (ssh != null) {
-                    ssh.disconnect();
-                }
-            } catch (IOException e) {
-                logger.warn("Error closing ssh connection.", e);
```
rkondratowicz: looks like behaviour is now changed.
before: logging a warning
now: throws `FtpException`.

Was this intended?

src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/FtpClient.java
```diff
@@ -204,4 +192,11 @@ private boolean isReportFile(RemoteResourceInfo resourceInfo) {
             && resourceInfo.getName().toLowerCase(Locale.getDefault()).endsWith(".csv");
     }
 
+    private String getCallerName(Exception exception) {
+        return Arrays.stream(exception.getStackTrace())
+            .skip(1) // first one should be `runWith`
+            .findFirst()
+            .map(StackTraceElement::getMethodName)
+            .orElse("UnknownMethodName");
+    }
```
rkondratowicz: Can you explain how that would help us?

src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/FtpClient.java
doncem: Closing ssh client is basically a failure. This code was, my guess, done because it was not closeable in the first place. Hence

```java
SSHClient client = null;

try {
    ...
} catch (IOException exc) {
    try closing
}
```

any IOException before closing is still caught and re-thrown with correct/better name inside

rkondratowicz: Throwing an exception can result in rolling back a transaction for example.
I assume the exception from closing connection was not rethrown and instead just logged as warning before for a reason.

doncem: Made compatible with https://github.com/hmcts/send-letter-service/pull/370/commits/8942b7851060a9aeed4ff3ea94021e1fa8865c90. It should've used try with resources from the beginning but since it is the topic, making necessary changes now. Didn't check but could have been not yet `Closeable` at the time of usage

src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/SshClient.java
```diff
@@ -0,0 +1,31 @@
...
+import net.schmizz.sshj.SSHClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.function.Function;
+
+/**
+ * Class to manage {@link SSHClient} with current state of service.
+ * For backwards compatibility the overridden {@link SshClient#close()}
```
lgonczar: This class is a bit confusing. All it does is catching an exception when closing fails - something that should be best done outside of it, anyway (so that the caller of the tool knows).

Does backward compatibility apply if we're also changing the code that uses the SSH client?

In my opinion would be better not to have this class.

doncem: So it was perfectly fine before? Fine

PR #371, opened 2019-04-24
Improve logging when marking letters as posted

rkondratowicz:
Minor changes:
- add extra info to logs
  - log which file we start going over and how many letters it has
  - include file name if errors/warnings
- if letter is not found in DB, log that fact as *error* (was 'warning') https://tools.hmcts.net/jira/browse/BPS-608

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
@@ -69,7 +69,15 @@ public class MarkLettersPostedTask {
                 .map(parser::parse)
                 .forEach(parsedReport -> {
                     insights.trackPrintReportReceived(parsedReport);
-                    parsedReport.statuses.forEach(this::markAsPosted);
+                    logger.info(
+                        "Updating letters from report {}. Letter count: {}",
+                        parsedReport.path,
+                        parsedReport.statuses.size()
+                    );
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
```diff
@@ -69,7 +69,15 @@ public void run() {
                 .map(parser::parse)
                 .forEach(parsedReport -> {
                     insights.trackPrintReportReceived(parsedReport);
-                    parsedReport.statuses.forEach(this::markAsPosted);
+                    logger.info(
+                        "Updating letters from report {}. Letter count: {}",
+                        parsedReport.path,
+                        parsedReport.statuses.size()
+                    );
```
doncem: this duplicates previous tracking via insights. maybe we should just use one?

rkondratowicz: I assumed the custom events are for different purpose.
This new info entry makes it easy to see what happened step by step in MarkLettersPostedTask by running a simple query.

doncem: to my knowledge no1 yet exploiting the custom event tracked via insights. might as well just use info trace alone - there's no need to duplicate data

PR #373, opened 2019-04-25
Include `ApacheHttpClient` for feign connections

doncem:
### Change description ###

Additional bean for `feign.Client` is apparently required for application insights' agent. Purposefully skipped during #369 😞 

```
[ ] Yes
[x] No
```


timja:
@nitinprabhu you can point your AKS at an app insights instance...

timja:
also if you do want app insights per pr you can do that... not sure if you would want that though

PR #377, opened 2019-04-26
Update use of ftp upload task

doncem:

[Performance - Prod shows degradation with FtpClient request](https://tools.hmcts.net/jira/browse/BPS-624)

### Change description ###

The tracking of upload was incorrect to begin with. Before:

```java
void upload(...) {
    Instant start = now();
    runWith(sftpClient -> {
        // actually does the upload and time tracking
    });
}
```

where `runWith` function connects to sftp, creates a client and then executes the actual action. This is per each file upload 😱.

This PR tears apart such bond by splitting as follows:

```java
ftpClient.runWith(sftpClient -> {
    ftpClient.upload(...);
    return anything;
}
```

Giving such construct you can provide any lambda to feed the function in `runWith` step. Since we have a loop in upload task which gets next 10 letters it is not too much but still should be a significant improvement. With this approach we have single connection per one batch in the parent loop aka max 10 letters can be sent within one sftp connection. Furthermore, lambda has more strict requirements for fields therefore inspect how it has changed more carefully if necessary.

Summary:

- Super beneficial accurate reporting on how long it took to upload files 💯
- Less connecting to ftp to do the job
- Extra trace log for each uploaded file to gather stats in more efficient way via appinsights

```
[ ] Yes
[x] No
```


src/integrationTest/java/uk/gov/hmcts/reform/sendletter/services/FtpUploadTest.java
```diff
@@ -24,7 +24,10 @@ void uploads_file() throws Exception {
         FileToSend doc = new FileToSend("hello.zip", "world".getBytes(), false);
         try (LocalSftpServer server = LocalSftpServer.create()) {
             FtpClient client = FtpHelper.getSuccessfulClient(LocalSftpServer.port);
-            client.upload(doc, LocalSftpServer.SERVICE_FOLDER);
+            client.runWith(sftpClient -> {
```
nitinprabhu: I would suggest to rename FtpClient -> FtpHelper or FtpUtil. Otherwise it is causing confusion while reading the code looks like FTP client run with take sftp client

src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/FtpClient.java
```diff
@@ -49,36 +49,34 @@ public FtpClient(
...
-                    "File uploaded. Time: {}, Size: {}, Folder: {}",
-                    ChronoUnit.MILLIS.between(start, Instant.now()) + "ms",
-                    file.content.length / 1024 + "KB",
-                    serviceFolder
-                );
+            isSuccess = true;
+        } catch (IOException exc) {
+            throw new FtpException("Unable to upload file.", exc);
+        } finally {
+            long millis = ChronoUnit.MILLIS.between(start, Instant.now());
```
nitinprabhu: Nice calculating it only once 👍

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/services/FtpUploadTest.java
doncem: Some of those tests are cramming actual integration testing purpose - they are there because they are there ¯\\\_(ツ)_/¯. I wouldn't want to move/rename them in this scope as it was already a fundamental change. Since not breaking - best to leave as a separate stand-alone review, sounds good?

src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/FtpClient.java
```diff
@@ -49,36 +49,34 @@ public FtpClient(
...
+            insights.trackFtpUpload(Duration.ofMillis(millis), isSuccess);
 
-                return null;
-            } catch (IOException exc) {
-                throw new FtpException("Unable to upload file.", exc);
-            } finally {
-                insights.trackFtpUpload(Duration.between(start, Instant.now()), isSuccess);
-            }
-        });
+            logger.info(
```
lgonczar: I can see it was already like this before this PR, so I'm fine with it staying as it is, but it would be worth loggin errors in case of failures at some point here.

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -63,18 +63,24 @@ public void run() {
...
-                .forEach(letter -> {
-                    uploadToFtp(letter);
-                    markAsUploaded(letter);
-                });
-            counter += letters.size();
-        } while (!letters.isEmpty());
+            uploaded = counter;
+            // applying single connection per single loop run.
+            // should be a significant speed improvement already reported
+            counter += ftp.runWith(sftpClient ->
```
lgonczar: Since our goal here is to address the problem of opening the connection each time we want to upload a file, why not solve it entirely and only open it once every run of the task?

doncem: this might introduce sudden connection drops especially after downtime window when theoretically letters pile up. or whenever there are loads of requests done during some point of the day. the approach is start small and review. the suggestion of yours is a good candidate for next pr along with _keep alive_ profile for ssh client

lgonczar: When we open the connection once per run, then even if that connection is broken, the files will get uploaded within a new connection, in the next run, which will be in 30 seconds. Given that opening a connection once per job run seems to be about as much of a change as this, I don't see why we shouldn't do it in the first go.

lgonczar: To be clear, I'm not blocking the PR with this comment

lgonczar: Another improvement, not necessarily in this PR, but still something we should do, would be not opening a connection if there's nothing to send.

src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/FtpClient.java
lgonczar: Having another look, I think this is the only thing we'll log if we fail to upload a file. As a result, we'll have no errors logged. Doesn't have to be addressed in this PR and it's nothing you've done, but needs to be handled at some point.

src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/FtpClient.java
rkondratowicz: info on successful upload (below) is now logged even on failures 👎

src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/FtpClient.java
rkondratowicz: This used to be logged in `try` block, move back. We don't want this in case of failures.

doncem: Failure logging is automatically done by app insights by pushing exception telemetry which has not been caught aka failure case is logged. This log entry is informational analog of dependency tracking but with finer details in it. Slightly amended the message to include yes/no and full path where it was attempted to go instead of just folder. Exception message includes full path too so it can be tracked back to this informational material

lgonczar: The info message says `File uploaded: ...`, which may actually mean `Failed to upload file`. While there is a flag later in the message that says whether it's a success of failure, it's going to be very easy to misinterpret it. I'm a huge fan of info saying `File uploaded` and error saying `Error when uploading file` - no way to get those messages wrong.

nitinprabhu: I am fine with both
1. Either log separately like @lgonczar suggested or
2. Update log message to say File uploaded successfully ? Yes ...

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -63,27 +64,39 @@ public void run() {
...
-                    markAsUploaded(letter);
-                });
-            counter += letters.size();
-        } while (!letters.isEmpty());
-
-        if (counter > 0) {
-            insights.trackUploadedLetters(counter);
+        List<Letter> lettersToUpload = repo.findFirst10ByStatus(LetterStatus.Created);
+
+        if (!lettersToUpload.isEmpty()) {
```
lgonczar: The way it works is pretty optimal - we read letters once and we open the connection only when there's something to upload.

How about:
```
while (!lettersToUpload.isEmpty()) {
   ...
   lettersToUpload = repo.findFirst10ByStatus(LetterStatus.Created);
}
```

This way we'd get rid of `letters` and some other bits of code.

doncem: `lettersToUpload` will become dynamic value and lambda function will not appreciate that. Would be nice to have some streaming workaround but I think it is only available from Java 9. On the other hand it will still require to fetch first batch in order to check whether worth opening ftp connection or not.

lgonczar: Right, I forgot that it's wrapped in a function.

src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/FtpClient.java
rkondratowicz: Keep the changes around single connection, please.
Feel free to introduce other changes in separate PR(s).

PR #381, opened 2019-04-30
Update AppInsights dependency types

doncem:
### Change description ###

This will make dependency data properly typed + application map feature in azure portal will look nicer

```
[ ] Yes
[x] No
```


doncem:
Not sure what you were after (and mostly - why?), but here it is, sample:
sprod instrumentation

can't get more info but it's a trivial change - just instead of "OTHER" show actual types

PR #384, opened 2019-04-30
Create a method in letter repository to find stale letters

lgonczar:

### Change description ###

Create a method in letter repository to find stale letters. It will be called in a future PR within current story.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/entity/StaleLetterSearchTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/entity/StaleLetterSearchTest.java
@@ -1,0 +1,100 @@
+package uk.gov.hmcts.reform.sendletter.entity;
+
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.jdbc.AutoConfigureTestDatabase;
+import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest;
+import uk.gov.hmcts.reform.sendletter.SampleData;
+
+import java.time.LocalDateTime;
+import java.util.List;
+
+import static java.util.stream.Collectors.toList;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
+@DataJpaTest
+public class StaleLetterSearchTest {
+
+    private static final String SMOKE_TEST_LETTER_TYPE = "smoke_test";
+
+    @Autowired
+    private LetterRepository repository;
+
+    @Test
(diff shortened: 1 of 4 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/sendletter/entity/StaleLetterSearchTest.java
```diff
@@ -0,0 +1,100 @@
...
+        assertThat(letters).isEmpty();
+    }
+
+    @Test
+    public void should_not_include_smoke_test_letters_in_the_result() {
+        // given
+        LocalDateTime cutOffDate = LocalDateTime.now().minusDays(2);
+
+        Letter smokeTestLetter = storeLetter(cutOffDate.minusDays(1), LetterStatus.Created, SMOKE_TEST_LETTER_TYPE);
+        storeLetter(cutOffDate.minusDays(1), LetterStatus.Created,"not-smoke-test");
```
rkondratowicz: missing space 😱

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
```diff
@@ -25,6 +25,14 @@
         @Param("before") LocalDateTime before
     );
 
+    @Query("select l from Letter l where l.status not in ('Posted', 'Aborted')"
+        + " and l.createdAt < :createdBefore and l.type <> '"
```
doncem: SLA really applies to `sentToPrintAt` value as it is _exact_ figure when external service is aware of letter supposed to be printed. Plus downtime window - shifts the business day really.

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterStatus.java
```diff
@@ -1,7 +1,15 @@
 package uk.gov.hmcts.reform.sendletter.entity;
 
+/**
+ * Represents possible letter statuses in the DB.
+ * <p>
+ * Note that in the repository these statuses are referenced by value, so any change to these names
+ * must be reflected in those queries.
+ * </p>
+ */
```
doncem: 👍. Yep, because of static value

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
lgonczar: A stale letter is now a letter that hasn't been posted, even though it's been 2 business days since receiving the request. It's not just about the external service's SLA anymore - we want to be alerted whenever for any reason a letter wasn't posted, regardless of where it got stuck.

As for the date, it's not the repository's concern (it receives it as a parameter), but we should definitely take another look at https://tools.hmcts.net/jira/browse/BPS-621 and make it clear when a business day starts - 00:00 or 17:00.

PR #391, opened 2019-05-02
Add code to calculate date that was X weekdays ago

lgonczar:

### Change description ###

Add code to calculate date that was X weekdays ago. This code is intended to be modified so that it takes bank holidays into account.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/services/date/DateService.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/services/date/DateService.java
@@ -1,0 +1,27 @@
+package uk.gov.hmcts.reform.sendletter.services.date;
+
+import org.springframework.stereotype.Service;
+
+import java.time.ZonedDateTime;
+
+@Service
+public class DateService {
+
+    public ZonedDateTime subtractBusinessDays(ZonedDateTime dateTime, int numberOfBusinessDays) {
+        int daysSubtracted = 0;
+        ZonedDateTime adjustedDateTime = dateTime;
+
+        while (daysSubtracted < numberOfBusinessDays) {
+            adjustedDateTime = adjustedDateTime.minusDays(1);
+            if (isBusinessDay(adjustedDateTime)) {
+                daysSubtracted++;
+            }
+        }
+
+        return adjustedDateTime;
+    }
+
+    private boolean isBusinessDay(ZonedDateTime dateTime) {
+        return dateTime.getDayOfWeek().getValue() < 6;
+    }
+}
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/services/date/DateService.java
```diff
@@ -0,0 +1,27 @@
...
+import java.time.ZonedDateTime;
+
+@Service
+public class DateService {
+
+    public ZonedDateTime subtractBusinessDays(ZonedDateTime dateTime, int numberOfBusinessDays) {
+        int daysSubtracted = 0;
+        ZonedDateTime adjustedDateTime = dateTime;
+
+        while (daysSubtracted < numberOfBusinessDays) {
```
lgonczar: We could calculate the day instead of iterating, but when we start taking bank holidays into account, that approach would be more complicated and given that we're only going to subtract a couple days, not worth it.

src/main/java/uk/gov/hmcts/reform/sendletter/services/date/DateService.java
```diff
@@ -0,0 +1,27 @@
...
+            if (isBusinessDay(adjustedDateTime)) {
+                daysSubtracted++;
+            }
+        }
+
+        return adjustedDateTime;
+    }
+
+    private boolean isBusinessDay(ZonedDateTime dateTime) {
+        return dateTime.getDayOfWeek().getValue() < 6;
```
nitinprabhu:
```suggestion
       return dateTime.getDayOfWeek().getValue() < DayOfWeek.SATURDAY.getValue();
```

src/main/java/uk/gov/hmcts/reform/sendletter/services/date/DateService.java
```diff
@@ -0,0 +1,27 @@
+package uk.gov.hmcts.reform.sendletter.services.date;
+
+import org.springframework.stereotype.Service;
+
+import java.time.ZonedDateTime;
+
+@Service
+public class DateService {
```
nitinprabhu:
```suggestion
public class DateCalculator {
```

src/main/java/uk/gov/hmcts/reform/sendletter/services/date/DateService.java
lgonczar: 👍

src/main/java/uk/gov/hmcts/reform/sendletter/services/date/DateService.java
lgonczar: Had problems with the previous date, too

PR #394, opened 2019-05-02
Add email sending feature

doncem:

[Automate send letter daily report](https://tools.hmcts.net/jira/browse/BPS-627)

### Change description ###

Introducing email sender component as stand-alone feature. No knowledge of contents/reports or any body if necessary. Brings in wide enough usage samples in case report does not need attachment but just some html body in there - can be provided by report separately. Or not report at all.

Ideally this should be a tiny library on it's own as similar implementation is in bulk scan project and potentially can be in others due to the nature of implementation introduced here

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/tasks/reports/EmailSender.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/tasks/reports/EmailSender.java
@@ -4,6 +4,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.mail.MailException;
 import org.springframework.mail.javamail.JavaMailSender;
 import org.springframework.mail.javamail.MimeMessageHelper;
 import org.springframework.stereotype.Component;
@@ -52,7 +53,7 @@ public class EmailSender {
             }
 
             mailSender.send(message);
-        } catch (MessagingException exc) {
+        } catch (MessagingException | MailException exc) {
             log.error("Error sending report", exc);
         }
     }
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/tasks/reports/EmailSender.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/tasks/reports/EmailSender.java
@@ -1,0 +1,75 @@
+package uk.gov.hmcts.reform.sendletter.tasks.reports;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.mail.javamail.JavaMailSender;
+import org.springframework.mail.javamail.MimeMessageHelper;
+import org.springframework.stereotype.Component;
+
+import javax.mail.MessagingException;
+import javax.mail.internet.MimeMessage;
+
+@Component
+@ConditionalOnProperty(prefix = "spring.mail", name = "host")
+public class EmailSender {
+
+    private static final Logger log = LoggerFactory.getLogger(EmailSender.class);
+
+    static final String EMAIL_BODY = "This is an auto generated email. Do not respond to it.";
+
+    private final JavaMailSender mailSender;
+    private final String from;
+
+    public EmailSender(
+        JavaMailSender mailSender,
+        @Value("${spring.mail.username}") String from
+    ) {
+        this.mailSender = mailSender;
+        this.from = from;
+    }
+
+    public void send(
+        String subject,
+        String body,
+        String[] recipients,
+        Attachment... attachments
+    ) {
+        try {
+            MimeMessage message = mailSender.createMimeMessage();
+
+            MimeMessageHelper helper = new MimeMessageHelper(message, true);
+
+            helper.setFrom(from);
+            helper.setTo(recipients);
+            helper.setSubject(subject);
+            helper.setText(body == null ? EMAIL_BODY : body);
+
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/reports/EmailSender.java
```diff
@@ -0,0 +1,75 @@
...
+public class EmailSender {
+
+    private static final Logger log = LoggerFactory.getLogger(EmailSender.class);
+
+    static final String EMAIL_BODY = "This is an auto generated email. Do not respond to it.";
+
+    private final JavaMailSender mailSender;
+    private final String from;
+
+    public EmailSender(
```
nitinprabhu: you have multiple send method which would be more useful if this was a library. May be just have one which we are going to use for now ?

doncem: I now. I was more hoping for general review and this being a non-breaking summary which I can easily remove. Plus will give me room to do an exception unit test case as well as it is left behind

src/test/java/uk/gov/hmcts/reform/sendletter/tasks/reports/EmailSenderTest.java
```diff
@@ -0,0 +1,132 @@
...
+    private static final String SUBJECT = "email subject";
+    private static final String RECIPIENT_A = "recipient A <a@localhost>";
+    private static final String RECIPIENT_B = "b@localhost";
+    private static final File ATTACHMENT_FILE;
+    private static final Attachment ATTACHMENT_1;
+    private static final Attachment ATTACHMENT_2;
+
+    static {
+        try {
+            ATTACHMENT_FILE = File.createTempFile("unit-test", "tst");
```
nitinprabhu: can we not have a file in resources folder for test instead of creating temp file for each run >

src/test/java/uk/gov/hmcts/reform/sendletter/tasks/reports/EmailSenderTest.java
```diff
@@ -0,0 +1,128 @@
...
+    private static final String TEST_PASSWORD = "test_password";
+
+    private static final String SUBJECT = "email subject";
+    private static final String RECIPIENT_A = "recipient A <a@localhost>";
+    private static final String RECIPIENT_B = "b@localhost";
+    private static final File ATTACHMENT_FILE;
+    private static final Attachment ATTACHMENT_1;
+    private static final Attachment ATTACHMENT_2;
+
+    static {
```
nitinprabhu: can we not put this in setUp method ?

src/test/java/uk/gov/hmcts/reform/sendletter/jupiter/GreenMailExtension.java
```diff
@@ -0,0 +1,66 @@
...
+import com.icegreen.greenmail.util.ServerSetupTest;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+/**
+ * This is a copy of not yet released code.
+ * {@see https://github.com/greenmail-mail-test/greenmail/tree/master/greenmail-junit5/src/main/java/com/icegreen/greenmail/junit5}
+ */
+public class GreenMailExtension extends GreenMailProxy implements BeforeEachCallback, AfterEachCallback {
```
nitinprabhu: Do we really need this ? can we not use `GreenMailRule` instead which we used in bulkscan ?

src/test/java/uk/gov/hmcts/reform/sendletter/tasks/reports/EmailSenderTest.java
doncem: just wanted to load once per class, not once per test case. not bothered too much. should i move to setup method?

src/test/java/uk/gov/hmcts/reform/sendletter/jupiter/GreenMailExtension.java
doncem: there's a docblock explaining this @nitinprabhu :D. right there one line above 😆

nitinprabhu: What I meant was we can use it once it is released ?

src/test/java/uk/gov/hmcts/reform/sendletter/tasks/reports/EmailSenderTest.java
nitinprabhu: We can use BeforeAll ?

src/test/java/uk/gov/hmcts/reform/sendletter/jupiter/GreenMailExtension.java
doncem: Yes, once it is released - definitely. Now we have to dump custom extension ourselves into codebase ;)

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/reports/EmailSender.java
```diff
@@ -0,0 +1,60 @@
...
+        this.from = from;
+    }
+
+    public void send(
+        String subject,
+        String[] recipients,
+        Attachment... attachments
+    ) {
+        try {
+            MimeMessage message = mailSender.createMimeMessage();
```
nitinprabhu: Adding info logs here would be nice- before and after the email has been sent

doncem: Yep, good point. Was thinking few moments ago about it. Just added a basic `debug` kind of log. The actual `info`s are awaiting on actual report

PR #387, opened 2019-05-01
Add service layer for the letters count report

aliveni:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/BPS-626

### Change description ###
Added service layer for the letters count report.
The Reports Service excludes the test service, adds a row with count "0" when no letters are uplaoded for the services.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[ x] No
```

PR #395, opened 2019-05-03
update envelope count summary repository to return stream

aliveni:
### Change description ###
Updated Reports repository to return `Stream`. 
(suggested in https://github.com/hmcts/send-letter-service/pull/387)

```
[ ] Yes
[x ] No
```


doncem:
i think there is always a benefit if you use unevaluated collections ;). disregarding the size of it - just a pointer to the top and then streaming over

lgonczar:
Given the size of the stream, it must be evaluated as soon as underlying data is requested. Otherwise, there would be too much overhead related with multiple roundtrips to the database.

doncem:
If I understand correctly, roundtrip will be executed for `ResultSet` > 10. Since we're only requesting 10 - there won't be any roundtrip present. There might as well be some config playing around. Nevertheless - not evaluating is always beneficial in this case. We know object, we just parse it, no extra expensive work is around there

lgonczar:
Effectively it's exactly the same as if we used a list. The collection is evaluated as soon as we need the data, just like a list would be. As I said before, no harm. Just no benefit, either.

doncem:
well. if you going to stream the list - there's a benefit for you right there. but disregard that, fine to evaluate....... .... ... and again stream all over. In my personal view in coding life there should never be an existing evaluated list unless you're finally sure. but we're java8 land yay. and still, best to never evaluate in report consensus

lgonczar:
We should look at what the specific case is about before applying generic rules to it. As soon as we call this method, we're sure we want the whole data. It's not a generic case. We're requesting the summary, which, even though it's a list, is an atomic piece of information to us - we want it as a whole and we want it when we request it. 

It's fine to stream when there's lots of data that we don't need in memory at once, for example, or for various optimisation reasons, or to build proxy streams, etc., but here we only want a tiny summary. As I said it doesn't hurt (so I approved the PR), but doesn't help with anything, either. We have other methods in the repository, where returning streams makes total sense, but I completely disagree that it's always better to use streams when retrieving data regardless of the context.

Also, as explained earlier, in our case the collection is evaluated anyway.

aliveni:
> > There is an error related with this PR:
> > ```
> > org.springframework.dao.InvalidDataAccessApiUsageException: You're trying to execute a streaming query method without a surrounding transaction that keeps the connection open so that the Stream can actually be consumed. Make sure the code consuming the stream uses @Transactional or any other way of declaring a (read-only) transaction.
> > ```
> > 
> > 
> > We could add `@Transactional` annotation to the service method, which would technically solve the problem, but would also leave us with an unwanted transaction. How about reverting this change?
> 
> Sure @lgonczar! I'm raising a pr to return `List` instead of `Stream`

@lgonczar Reverted the change in https://github.com/hmcts/send-letter-service/pull/410

aliveni:
> There is an error related with this PR:
> 
> ```
> org.springframework.dao.InvalidDataAccessApiUsageException: You're trying to execute a streaming query method without a surrounding transaction that keeps the connection open so that the Stream can actually be consumed. Make sure the code consuming the stream uses @Transactional or any other way of declaring a (read-only) transaction.
> ```
> 
> We could add `@Transactional` annotation to the service method, which would technically solve the problem, but would also leave us with an unwanted transaction. How about reverting this change?

Sure @lgonczar! I'm raising a pr to return `List` instead of `Stream`

PR #397, opened 2019-05-07
Create a service for retrieving stale letters

lgonczar:

### Change description ###

Create a service for retrieving stale letters from the database.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/services/StaleLetterService.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/services/StaleLetterService.java
@@ -50,7 +50,7 @@ public class StaleLetterService {
                 .toLocalDateTime()
         );
 
-        return mapToStaleLetters(dbLetters);
+        return dbLetters.map(this::mapToStaleLetter);
     }
 
     /**
@@ -83,16 +83,14 @@ public class StaleLetterService {
         );
     }
 
-    private Stream<StaleLetter> mapToStaleLetters(Stream<Letter> dbLetters) {
-        return dbLetters.map(letter ->
-            new StaleLetter(
-                letter.getId(),
-                letter.getStatus().name(),
-                letter.getService(),
-                letter.getCreatedAt(),
-                letter.getSentToPrintAt(),
-                letter.isFailed()
-            )
+    private StaleLetter mapToStaleLetter(Letter dbLetter) {
+        return new StaleLetter(
+            dbLetter.getId(),
+            dbLetter.getStatus().name(),
+            dbLetter.getService(),
+            dbLetter.getCreatedAt(),
+            dbLetter.getSentToPrintAt(),
+            dbLetter.isFailed()
         );
     }
 }
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/services/StaleLetterService.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/services/StaleLetterService.java
@@ -1,0 +1,98 @@
+package uk.gov.hmcts.reform.sendletter.services;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Service;
+import uk.gov.hmcts.reform.sendletter.entity.Letter;
+import uk.gov.hmcts.reform.sendletter.entity.LetterRepository;
+import uk.gov.hmcts.reform.sendletter.model.out.StaleLetter;
+import uk.gov.hmcts.reform.sendletter.services.date.DateCalculator;
+
+import java.time.Clock;
+import java.time.LocalTime;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.util.stream.Stream;
+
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/services/StaleLetterService.java
```diff
@@ -0,0 +1,98 @@
...
+                .truncatedTo(DAYS)
+                .plusNanos(ftpDowntimeAdjustedTime.toNanoOfDay());
+
+        return dateCalculator.subtractBusinessDays(
+            todayWithFtpDowntimeAdjustedTime,
+            minStaleLetterAgeInBusinessDays
+        );
+    }
+
+    private Stream<StaleLetter> mapToStaleLetters(Stream<Letter> dbLetters) {
```
doncem: `map` function should operate with objects only instead of stream of objects which becomes a `flatmap`

src/test/java/uk/gov/hmcts/reform/sendletter/services/StaleLetterServiceTest.java
```diff
@@ -0,0 +1,183 @@
...
+        int minStaleLetterAgeInBusinessDays = 123;
+
+        ZonedDateTime now = LocalDate.parse("2019-05-03")
+            .atTime(15, 59) // currently it's before FTP downtime
+            .atZone((ZoneId.of(EUROPE_LONDON)));
+
+        setCurrentTimeAndZone(EUROPE_LONDON, now);
+
+        // when
+        staleLetterService(ftpDowntimeStartTime, minStaleLetterAgeInBusinessDays).getStaleLetters();
```
doncem: missing stream closure

src/main/java/uk/gov/hmcts/reform/sendletter/services/StaleLetterService.java
```diff
@@ -0,0 +1,98 @@
...
+        return mapToStaleLetters(dbLetters);
+    }
+
+    /**
+     * Calculates the cut-off creation date for stale letters.
+     *
+     * @return Cut-off creation date for stale letters. Only letters older than that
+     *         can be considered stale.
+     */
+    private ZonedDateTime calculateCutOffCreationDate() {
```
rkondratowicz: let's say it's 14:00 now. Should a letter uploaded 2 days ago at 14:15 not be considered stale?

src/main/java/uk/gov/hmcts/reform/sendletter/services/StaleLetterService.java
lgonczar: I'm not sure what you're referring to. Are you saying the use of `.map` is incorrect or the method name is wrong?

src/main/java/uk/gov/hmcts/reform/sendletter/services/StaleLetterService.java
lgonczar: I clarified with the team what two business days mean and it's 48 hours within business days. Based on that, a letter uploaded 2 days ago at 14:15 will not be considered stale, as it's not two days old, yet.

src/main/java/uk/gov/hmcts/reform/sendletter/services/StaleLetterService.java
doncem: technically a function which accepts `Stream` and returns `Stream` is a flatMap. But you are actually returning same `Stream but mapped over hence the function is named `mapSomething`. I'm suggesting to keep the map but function should be direct conversion from `Letter` to `StaleLetter` and amend calling place slightly to call `stream.map(this::mapToStaleLetters)` instead ;)

lgonczar: `flatMap` is a combination of mapping and flattening - I don't see how this is happening here at all. Besides, we're not mapping a stream to stream, but letters to stale letters.

I'll do the direct mapping if that's your preference. It makes no difference from practical perspective.

doncem: i can read the method yes it is not flatmapping anything :). just interpreting method signature here. and besides, this was not a pr blocker - stream closures are

src/test/java/uk/gov/hmcts/reform/sendletter/services/StaleLetterServiceTest.java
lgonczar: This is a 100% in-memory stream, that lives in the scope of a single test. There's no need to close it, as it's guaranteed to be GC-collected. It's a similar situation to why you don't have to explicitly close a stream retrieved from an ArrayList.

It's crucial to close streams behind which there are unmanaged resources (this is the whole point of AutoClosable). For example, when consuming a stream retrieved from a JPA repository, we have to make sure it's always closed at the end. It's not the case here, though.

PR #396, opened 2019-05-03
Introduce daily report for summary of uploaded letters

doncem:
### JIRA link (if applicable) ###

[Automate send letter daily report](https://tools.hmcts.net/jira/browse/BPS-627)

### Change description ###

Incorporate #394 and #392 to send out an email. Which is disabled at the moment. All the changes are non-effective and left for final (next) PR.

This is another set of "purely additions" PR without breaking anything existing. 

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #399, opened 2019-05-07
Feature: enable daily report task

doncem:

[Automate send letter daily report](https://tools.hmcts.net/jira/browse/BPS-627)

### Change description ###

Going through configuration and infrastructure to enable report for prod only. The secrets are not yet created and will confirm once they are by removing the `DO NOT MERGE` label. Will not start creating secrets until PR is not approved as it will confirm that naming is acceptable.

This PR introduces config separation among different beans: both `EmailSender` and `DailyLetterUploadSummaryReport` was dependant on single config value - `spring.mail.host` which is only relevant for former. Report itself should not repeat this so providing 2 conditions: new flag plus condition on `EmailSender` bean being present (it is dependency).

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/reports/DailyLetterUploadSummaryReportTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/reports/DailyLetterUploadSummaryReportTest.java
@@ -1,0 +1,29 @@
+package uk.gov.hmcts.reform.sendletter.tasks.reports;
+
+import org.junit.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.mock.mockito.SpyBean;
+import org.springframework.mail.javamail.JavaMailSender;
+
+import javax.mail.internet.MimeMessage;
+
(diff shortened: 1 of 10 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/sendletter/tasks/reports/DailyLetterUploadSummaryReportTest.java
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.boot.test.mock.mockito.SpyBean;
+import org.springframework.mail.javamail.JavaMailSender;
+
+import javax.mail.internet.MimeMessage;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.verify;
+
+@SpringBootTest
+public class DailyLetterUploadSummaryReportTest {
```
nitinprabhu: should we also include test for when recipient list is not present ?

doncem: We have it from #396 as a unit test. Since these are cron based report tasks these integration tests purely stands for "kind of functional" ones, integrity verifiers.

doncem:
The secrets have been populated. Will merge in a bit

PR #404, opened 2019-05-07
Include bank holidays in business days calculator

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/sendletter/services/date/DateCalculator.java
```diff
@@ -1,28 +1,61 @@
...
-    private boolean isBusinessDay(ZonedDateTime dateTime) {
-        return dateTime.getDayOfWeek().getValue() < DayOfWeek.SATURDAY.getValue();
+    private boolean isBusinessDay(ZonedDateTime dateTime, List<LocalDate> bankHolidays) {
+        return dateTime.getDayOfWeek().getValue() < DayOfWeek.SATURDAY.getValue()
+            && !bankHolidays.contains(dateTime.toLocalDate());
+    }
+
+    private List<LocalDate> getBankHolidayDates() {
+        try {
+            return this.bankHolidaysClient
```
lgonczar: Not a part of this pull request, but maybe the client could be specific to England (it doesn't support any other option) and then it would have a method called `getHolidays` instead of `getForEngland`? This would simplify it. For example it could have the whole URL to the holidays endpoint in configuration.

src/main/java/uk/gov/hmcts/reform/sendletter/config/BankHolidaysConfig.java
```diff
@@ -0,0 +1,15 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.web.client.RestTemplate;
+import uk.gov.hmcts.reform.sendletter.services.date.holidays.BankHolidaysClient;
+
+@Configuration
+public class BankHolidaysConfig {
+
+    @Bean
+    public BankHolidaysClient client(RestTemplate restTemplate) {
+        return new BankHolidaysClient(restTemplate, " https://www.gov.uk/bank-holidays");
```
lgonczar: Best to have the URL in a config file. Also, it starts with a space here.

nitinprabhu: Yes even I thought of it 😀

src/main/java/uk/gov/hmcts/reform/sendletter/services/date/DateCalculator.java
rkondratowicz: done

src/main/java/uk/gov/hmcts/reform/sendletter/services/date/DateCalculator.java
```diff
@@ -1,28 +1,61 @@
...
     }
 
-    private boolean isBusinessDay(ZonedDateTime dateTime) {
-        return dateTime.getDayOfWeek().getValue() < DayOfWeek.SATURDAY.getValue();
+    private boolean isBusinessDay(ZonedDateTime dateTime, List<LocalDate> bankHolidays) {
+        return dateTime.getDayOfWeek().getValue() < DayOfWeek.SATURDAY.getValue()
+            && !bankHolidays.contains(dateTime.toLocalDate());
+    }
+
+    private List<LocalDate> getBankHolidayDates() {
```
lgonczar: This could even be a set, as all we care about is whether it contains certain dates. In practice, the difference would be next to none, though.

src/main/java/uk/gov/hmcts/reform/sendletter/services/date/holidays/BankHolidaysClient.java
```diff
@@ -1,19 +1,29 @@
...
 
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
 import org.springframework.web.client.RestTemplate;
 import uk.gov.hmcts.reform.sendletter.services.date.holidays.response.Holidays;
 
+@Component
 public class BankHolidaysClient {
 
+    private static final String DEFAULT_URL = "https://www.gov.uk/bank-holidays/england-and-wales.json";
```
lgonczar: I'd still recommend taking it from configuration, as that's where one would search for it.

rkondratowicz: taking a cue from gov notify client which has url hardcoded

lgonczar: We include our dependencies' URLs in the config by convention. This one makes an exception now.

PR #408, opened 2019-05-08
Added reports service config to map db service name to display name on reports

aliveni:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/BPS-638

### Change description ###
Added new config to map db service names to display names on report.
Updated reports service to use the new config.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x ] No
```

PR #410, opened 2019-05-08
Cleanup reports endpoint service

aliveni:
Addressed review comments from https://github.com/hmcts/send-letter-service/pull/408
Updated `LettersCountSummaryRepository` to return `List` instead of `Stream`.
Added integration test for report end point


src/main/java/uk/gov/hmcts/reform/sendletter/services/ReportsService.java
```diff
@@ -45,6 +46,7 @@ public ReportsService(
         this.timeToHour = downtimeFromHour;
     }
 
+    @Transactional
```
nitinprabhu: should we not have an integration test which will add dummy records to db as part of setup and retrieve records from db and verify?

aliveni: @nitinprabhu added integration test for the reports endpoint.

src/test/java/uk/gov/hmcts/reform/sendletter/SampleData.java
```diff
@@ -56,6 +57,15 @@ public static LetterWithPdfsRequest letterWithPdfsRequest() {
         );
     }
 
+    public static uk.gov.hmcts.reform.sendletter.entity.Letter letterEntityWithSentToPrintAt(
```
nitinprabhu: can we not reuse `letterEntity`method and then set `letterEntity` where you are calling this ?

src/main/java/uk/gov/hmcts/reform/sendletter/config/ReportsServiceConfig.java
```diff
@@ -2,20 +2,51 @@
...
 import java.util.Optional;
 
+import static java.util.stream.Collectors.toMap;
+
 @ConfigurationProperties(prefix = "reports")
 public class ReportsServiceConfig {
 
     private Map<String, String> serviceConfig;
 
+    public static class Mapping {
```
aliveni: @rkondratowicz We need this class. The config was loaded in a different format without this class
`Eg:` The below config is loaded as 2 entries
```
[0].service=some_service_name
[0].display-name=some service
```

src/test/java/uk/gov/hmcts/reform/sendletter/SampleData.java
aliveni: we can. I'll update

aliveni: done

src/main/java/uk/gov/hmcts/reform/sendletter/config/ReportsServiceConfig.java
rkondratowicz: It was not used in the setter before (or anywhere else)

src/main/java/uk/gov/hmcts/reform/sendletter/config/ReportsServiceConfig.java
```diff
@@ -2,20 +2,51 @@
...
+    }
+
     public Map<String, String> getServiceConfig() {
         return serviceConfig;
     }
 
-    public void setServiceConfig(Map<String, String> serviceConfig) {
-        this.serviceConfig = serviceConfig;
+    public void setServiceConfig(List<Mapping> mappings) {
+        this.serviceConfig = mappings
```
rkondratowicz: perhaps we can just keep the list of object with clear names instead of converting to a `Map<String, String>`?

PR #415, opened 2019-05-09
Extend lock duration for reporting job

rkondratowicz:
to prevent sending 2 emails (from 2 instances)

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/tasks/reports/DailyLetterUploadSummaryReport.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/tasks/reports/DailyLetterUploadSummaryReport.java
@@ -58,7 +58,7 @@ public class DailyLetterUploadSummaryReport {
         }
     }
 
-    @SchedulerLock(name = "daily-letter-upload-summary")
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/reports/DailyLetterUploadSummaryReport.java
```diff
@@ -58,7 +58,7 @@ public DailyLetterUploadSummaryReport(
         }
     }
 
-    @SchedulerLock(name = "daily-letter-upload-summary")
+    @SchedulerLock(name = "daily-letter-upload-summary", lockAtLeastFor = 5_000)
```
rkondratowicz:
```
    /**
     * The lock will be held at least for X millis. Can be used if you really need to execute the task
     * at most once in given period of time. If the duration of the task is shorter than clock difference between nodes, the task can
     * be theoretically executed more than once (one node after another). By setting this parameter, you can make sure that the
     * lock will be kept at least for given period of time.
     */
    long lockAtLeastFor() default -1;
```

PR #417, opened 2019-05-09
Log warning when stale letters are found by the scheduled job

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/StaleLettersTask.java
```diff
@@ -40,7 +40,12 @@ public void run() {
         logger.info("Started '{}' task", TASK_NAME);
 
         List<Letter> letters = staleLetterService.getStaleLetters();
-        long count = letters.stream().peek(insights::trackStaleLetter).count();
```
rkondratowicz: well hidden crucial operation... 😱
 
from javadoc on `peek`
```
This method exists mainly to support debugging, where you want to see the elements as they flow past a certain point in a pipeline
```

PR #429, opened 2019-05-17
Add endpoint for listing letters pending upload

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/sendletter/services/PendingLettersService.java
```diff
@@ -0,0 +1,26 @@
...
+public class PendingLettersService {
+
+    private final LetterRepository repo;
+
+    public PendingLettersService(LetterRepository repo) {
+        this.repo = repo;
+    }
+
+    public List<Letter> getPendingLetters() {
+        return repo
```
lgonczar: Based on this code I assume this repository method returns a stream. If that's the case, that stream has to be closed in every case (success or failure), so we need to wrap it in try-with-resources. There's a DB connection and transaction related with it.

lgonczar: Also, there will have to be a `@Transactional` annotation somewhere, to make it work.

src/test/java/uk/gov/hmcts/reform/sendletter/services/PendingLettersServiceTest.java
```diff
@@ -0,0 +1,29 @@
...
+import uk.gov.hmcts.reform.sendletter.entity.LetterStatus;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+class PendingLettersServiceTest {
+
+    @Test
+    void should_read_letters_in_proper_status_from_repo() {
```
lgonczar: We also need to verify the result of the method.

src/test/java/uk/gov/hmcts/reform/sendletter/services/PendingLettersServiceTest.java
```diff
@@ -0,0 +1,38 @@
...
+        // given
+        LetterRepository repo = mock(LetterRepository.class);
+        PendingLettersService service = new PendingLettersService(repo);
+        List<Letter> noLetters = Collections.emptyList();
```
lgonczar: Empty list is a good case to test 👍. It would also be useful to have a test for a non-empty list. Alternatively, if we just care if the service has returned what it got the repository, we can create an assertion for exactly the same instance.

doncem: did not want to tamper too much with existing pr but can include extra test just for the sake of validity

PR #438, opened 2019-05-23
Add cleanup cron variable to infra config

doncem:
### Change description ###

This PR introduces a change which has hidden agenda: force change app settings to refresh new secret stored in vault

```
[ ] Yes
[x] No
```


infrastructure/variables.tf
```diff
@@ -74,6 +74,11 @@ variable "file_cleanup_enabled" {
   description = "Determines whether old files should be deleted from the FTP server. Should only be enabled on AAT."
 }
 
+variable "file_cleanup_cron" {
+  default     = "0 0 5 * * *"
```
nitinprabhu: You have defined the default here which will cause to run remove files on prod ?

doncem: Default value is already present in main application config. Enable/Disable flag is above and it will be disabled for prod as it now

doncem:
Can always kick in #437 as a second deployment. It doesn't need 2 deployments as it is only moving one code to another place

doncem:
few PRs already got in, can merge this one as well

PR #443, opened 2019-06-09
Updated finger print as supplier has updated server

nitinprabhu:
### Change description ###

- Third party supplier has updated their server as a result the finger print which we store needs to be updated.

-  Before we merge this PR could we please confirm the new generated fingerprint is correct.

```
[ ] Yes
[x ] No
```


nitinprabhu:
Merging this as it works on AAT with the updated fingerprint

PR #449, opened 2019-06-11
Close SSHClient (and not SFTPClient) after file transfer

lgonczar:
IGNORE, DO NOT MERGE

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/sendletter/ProcessMessageTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/sendletter/ProcessMessageTest.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.reform.sendletter;
 
+import net.schmizz.sshj.SSHClient;
 import net.schmizz.sshj.sftp.RemoteFile;
 import net.schmizz.sshj.sftp.RemoteResourceInfo;
 import net.schmizz.sshj.sftp.SFTPClient;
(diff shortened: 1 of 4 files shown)


src/functionalTest/java/uk/gov/hmcts/reform/sendletter/ProcessMessageTest.java
```diff
@@ -31,7 +32,8 @@ void should_send_letter_and_upload_file_on_sftp_server_when_letter_contains_one_
             sampleLetterRequestJson("letter_single_document.json", "two-page-template.html")
         );
 
-        try (SFTPClient sftp = getSftpClient()) {
+        try (SSHClient ssh = getSshClient()) {
+            SFTPClient sftp = ssh.newSFTPClient();
```
nitinprabhu: We can do this inside the `waitForFileOnSftp` so that we don't need to repeat `ssh.newSFTPClient();` ?

rkondratowicz: Half of this class is highlighted in yellow in intellij, there's so much repetition already on master.
Perhaps something to do address on a separate occasion.

lgonczar: This is a quick fix rather than anything else. Lot's of code in this area is (and has been) duplicated, which is not ok, but let's treat it as a separate issue.

lgonczar: > We can do this inside the waitForFileOnSftp so that we don't need to repeat ssh.newSFTPClient(); ?

The sftp client is used later in these `try` statements. There are ways to optimise this, but perhaps we should leave it for the deduplication story.

PR #458, opened 2019-06-18
Change FTP file cleanup cron, so that it runs every hour

lgonczar:
### Change description ###

Change FTP file cleanup cron, so that it runs every hour. It doesn't delete files newer than 12h, anyway.

```
[ ] Yes
[x] No
```


lgonczar:
> Although it'll do the trick, but do we want to keep the same in `application.yml` too?

makes sense

PR #472, opened 2019-06-27
Speed up marking letters as posted

rkondratowicz:
no need to load the entire content of a file for each letter

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
@@ -35,7 +35,7 @@ public interface LetterRepository extends JpaRepository<Letter, UUID> {
     @Query("select l.status from Letter l where l.id = :id")
     Optional<LetterStatus> findStatusById(@Param("id") UUID id);
 
-    @Modifying
+    @Modifying(clearAutomatically = true)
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
```diff
@@ -31,6 +32,17 @@
 
     Optional<Letter> findById(UUID id);
 
+    @Query("select l.status from Letter l where l.id = :id")
+    Optional<LetterStatus> findStatusById(@Param("id") UUID id);
+
+    @Modifying(clearAutomatically = true)
```
rkondratowicz:
```
	/**
	 * Defines whether we should clear the underlying persistence context after executing the modifying query.
	 *
```

rkondratowicz: without this, in tests, an old version of letter is fetched:
```
repository.saveAndFlush(letter);
// run the task
letter = repository.findById(letter.getId()).get(); // this letter was still in 'posted' status
```

nitinprabhu: Can we not use @DirtiesContext on the test class ?

PR #473, opened 2019-06-28
Fix class name in logging an unhandled exception

lgonczar:
### Change description ###

Fix class name in logging an unhandled exception

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/sendletter/config/ThreadPoolConfig.java
```diff
@@ -46,7 +46,7 @@ public TaskScheduler sendLetterTaskScheduler() {
         ThreadPoolTaskScheduler taskScheduler = new SendLetterTaskScheduler();
         taskScheduler.setThreadNamePrefix("SendLetterTask-");
         taskScheduler.setErrorHandler(t -> {
-            log.error("Unhandled exception during task. {}: {}", t.getClass(), t.getMessage(), t);
+            log.error("Unhandled exception during task. {}: {}", t.getClass().getName(), t.getMessage(), t);
```
doncem: `getName` or `getSimpleName`?

doncem:
Re-opening as it looks like a good fix to me. Please close again if you disagree

rkondratowicz:
can this be merged @lgonczar

lgonczar:
I created this PR based on a false observation. Class has a proper `toString` implementation, so I came to the conclusion that this PR was unnecessary.

doncem:
Yes, classes usually get serialised with package info as would `.getName()` I believe. Doesn't hurt to use actual method rather than trusting some serialiser 🙂. Happy with this method or `getSimpleName()`. Close PR with reason why so it's not reopened

PR #475, opened 2019-07-02
Update scheduled jobs config format

rkondratowicz:
in preparation for adding a new property for 'upload-letters' job that will indicate which fingerprint should be used when searching for letters to upload.

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -104,9 +104,12 @@ scheduling:
   lock_at_most_for: ${SCHEDULING_LOCK_AT_MOST_FOR:PT10M} # 10 minutes in ISO-8601
 
 tasks:
-  upload-letters-interval-ms: ${UPLOAD_LETTERS_INTERVAL:30000}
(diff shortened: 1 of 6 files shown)


src/main/resources/application.yaml
```diff
@@ -104,9 +104,12 @@ scheduling:
   lock_at_most_for: ${SCHEDULING_LOCK_AT_MOST_FOR:PT10M} # 10 minutes in ISO-8601
 
 tasks:
-  upload-letters-interval-ms: ${UPLOAD_LETTERS_INTERVAL:30000}
-  mark-letters-posted: ${FTP_REPORTS_CRON:0 30 * * * *}
-  stale-letters-report: ${STALE_LETTERS_REPORT_CRON:0 30 11 * * *}
+  upload-letters:
+    interval-ms: ${UPLOAD_LETTERS_INTERVAL:30000}
```
rkondratowicz: i'll add something like `key-fingerprint` here (along with the code change) on next PR

PR #476, opened 2019-07-02
Filter letters by fingerprint when uploading

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
```diff
@@ -15,6 +15,8 @@
 
     List<Letter> findFirst3ByStatus(LetterStatus status);
 
+    List<Letter> findFirst3ByStatusAndEncryptionKeyFingerprint(LetterStatus status, String fingerprint);
```
aliveni: IMO the method name can be shorted

rkondratowicz: note that the SQL query is autogenerated by JPA here, it's based on the method name.

PR #477, opened 2019-07-02
Update the diagram

rkondratowicz:
The current one is huge, takes an entire page of readme and contains (imho) too much details.
Let me know what you think.

You can view the new version here:
https://github.com/hmcts/send-letter-service/blob/refactor/update-diagram/README.md


rkondratowicz:
added readme on how to edit it

PR #479, opened 2019-07-04
Enable encryption on AAT

rkondratowicz:
note that the key is already set on AAT, only feature toggle needs switching

### DIFF
--- a/infrastructure/aat.tfvars
+++ b/infrastructure/aat.tfvars
@@ -2,4 +2,6 @@ vault_section = "preprod"
 scheduling_enabled = "true"
 file_cleanup_enabled = "true"
 
+encyption_enabled = "true"
+
 capacity = "2"


infrastructure/aat.tfvars
```diff
@@ -2,4 +2,6 @@ vault_section = "preprod"
 scheduling_enabled = "true"
 file_cleanup_enabled = "true"
 
+encyption_enabled = "true"
```
aliveni: @rkondratowicz just noticed it is misspelled

rkondratowicz: well spotted!
It's misspelled everywhere though (see https://github.com/hmcts/send-letter-service/blob/master/infrastructure/variables.tf#L64) so this should work.
I'll fix the spelling everywhere separately.

PR #492, opened 2019-07-18
Update test dependencies

doncem:
### Change description ###

Over time few dependencies were left behind. Recent merge #490 did not include `scp`. Perhaps because dependency manager has enlisted as previous version being latest.

`Rest-assured` has been postponed until manual intervention. Need to register version with extension to make dependencies properly inject.

Another recent development occurred: sonarqube deprecations

Lastly, pitest. Not a frequent tool but it still exists somewhere and, perhaps, QA is using it. Or can use it

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -186,11 +186,12 @@ def versions = [
   mockitoJupiter: '2.27.0',
   pdfbox: '2.0.15',
   reformLogging: '5.0.1',
-  restAssured: '3.3.0',
   shedlock: '2.5.0',
   springBoot: plugins.getPlugin('org.springframework.boot').class.package.implementationVersion,
   springfoxSwagger: '2.9.2'
 ]
+
+ext["rest-assured.version"] = '4.0.0'
 
 dependencyManagement {
   dependencies {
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -182,16 +182,18 @@ repositories {
...
+  mockitoJupiter: '3.0.0',
   pdfbox: '2.0.15',
   reformLogging: '5.0.1',
-  restAssured: '3.3.0',
   shedlock: '2.5.0',
   springBoot: plugins.getPlugin('org.springframework.boot').class.package.implementationVersion,
   springfoxSwagger: '2.9.2'
 ]
 
+ext["rest-assured.version"] = '4.0.0'
```
rkondratowicz: why exactly does it have to be done this way?

doncem: assures all dependencies are imported with the same version. the dependencies are:

- `io.rest-assured:json-path`
- `io.rest-assured:rest-assured-common`
- `io.rest-assured:xml-path`

otherwise **only** `io.rest-assured:rest-assured` will be pulled in as version 4, those ^ - version 3

PR #500, opened 2019-07-24
Don't load full letter when marking as `Posted`

doncem:

[Optimise marking letters as posted](https://tools.hmcts.net/jira/browse/BPS-690)

### Change description ###

Initial PR: #472. Not sure why it was closed as no description left. Re-attempting to put code back in with minimal change possible

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
@@ -1,6 +1,7 @@
 package uk.gov.hmcts.reform.sendletter.entity;
 
 import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Modifying;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.query.Param;
 import uk.gov.hmcts.reform.sendletter.tasks.UploadLettersTask;
@@ -32,6 +33,13 @@ public interface LetterRepository extends JpaRepository<Letter, UUID> {
     Optional<Letter> findByChecksumAndStatusOrderByCreatedAtDesc(String checksum, LetterStatus status);
 
     Optional<Letter> findById(UUID id);
+
+    @Query("SELECT l.status FROM Letter l WHERE l.id = :id")
+    Optional<LetterStatus> findLetterStatus(@Param("id") UUID id);
+
+    @Modifying(clearAutomatically = true)
+    @Query("UPDATE Letter l SET l.status = 'Posted', l.printedAt = :printedAt, l.fileContent = null WHERE l.id = :id")
+    int markLetterAsPosted(@Param("id") UUID id, @Param("printedAt") LocalDateTime printedAt);
 
     Optional<Letter> findByIdAndService(UUID id, String service);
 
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
```diff
@@ -33,6 +34,13 @@
 
     Optional<Letter> findById(UUID id);
 
+    @Query("SELECT l.status FROM Letter l WHERE l.id = :id")
+    Optional<LetterStatus> findLetterStatus(@Param("id") UUID id);
+
+    @Modifying(clearAutomatically = true)
+    @Query("UPDATE Letter l SET l.status = 'Posted', l.printedAt = :printedAt, l.fileContent = null WHERE l.id = :id")
```
lgonczar: Theoretically, we could even try `... where l.id in (...)```, but this is still a huge improvement.

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
```diff
@@ -93,20 +92,20 @@ public void run() {
     }
 
     private void markAsPosted(LetterPrintStatus letterPrintStatus, String reportFileName) {
-        Optional<Letter> optional = repo.findById(letterPrintStatus.id);
+        Optional<LetterStatus> optional = repo.findLetterStatus(letterPrintStatus.id);
```
lgonczar: There's no need to query the DB for the status first. Instead, we can just update the record (`... where id = ... and status == 'Uploaded'`. If the query returns 1 (which is approx 100% cases), done. If not, then we can read the record and log the necessary details.

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
lgonczar: Never mind

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/MarkLettersPostedTask.java
doncem: This change assimilates the current behaviour. Would be way better to do bulk update on such task like printing flag update. Would even consider skipping the extra branch reporting on bad status and keep it for stale letter report instead

lgonczar: Fair enough

doncem: Shame, information where `printed_at` exists is retrieved from report file. So.. per letter :(

src/main/java/uk/gov/hmcts/reform/sendletter/services/LetterDataAccessService.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public LetterDataAccessService(LetterRepository repository) {
+        this.repository = repository;
+    }
+
+    public Optional<LetterStatus> findLetterStatus(UUID id) {
+        return repository.findLetterStatus(id);
+    }
+
+    @Transactional
```
lgonczar: Is it the requirement of the framework to call @Modifying methods in trasactions?

doncem: Yes. `@DataJpaTest`s did not fail because they are automatically marked as `@Transactional`. Only `EndToEnd` tests have full cover of app running "naturally". Personally I used to fail to get them running successfully on my machine so stopped checking. But it looks like has fixed itself - verified

PR #507, opened 2019-08-01
Remove FTP availability checker from tasks

doncem:
### Change description ###

Replace with annotation which is dealt in scheduled wrapped task - present already for all schedules. Just execute check in case annotation is present.

- Transparency in tasks
- Single point of check
- Less mockery in tests

P.S. Maybe could use `@FtpDowntime` annotation name instead of `@FtpConstraint` 🤔

Important reminder! This is suggestion PR. Review/Approve as you will and if thumbed up - can proceed further from here

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
@@ -13,18 +13,26 @@ import org.aspectj.lang.ProceedingJoinPoint;
 import org.aspectj.lang.annotation.Around;
 import org.aspectj.lang.annotation.Aspect;
 import org.aspectj.lang.annotation.Pointcut;
+import org.aspectj.lang.reflect.MethodSignature;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/logging/AppInsights.java
```diff
@@ -43,14 +51,30 @@
...
     public void aroundSchedule(ProceedingJoinPoint joinPoint) throws Throwable {
+        Method method = ((MethodSignature) joinPoint.getSignature()).getMethod();
+        FtpConstraint ftpConstraint = AnnotationUtils.findAnnotation(method, FtpConstraint.class);
+
+        if (ftpConstraint != null) {
+            if (!ftpAvailabilityChecker.isFtpAvailable(LocalTime.now(ZoneId.of(EUROPE_LONDON)))) {
+                log.info("Not processing '{}' task due to FTP downtime window", ftpConstraint.name());
+                return;
+            }
+        }
```
rkondratowicz: Why is checking FTP availability a part of logging component?

To sum up this PR:
Fancy? Yes.
Clearer, better? In my opinion, no.

👎 from me.

PR #514, opened 2019-08-18
Update deprecated chart template usage

doncem:
### Change description ###

Usage of `values.template.yaml` is deprecated and TTL is until the week after

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/rpe-send-letter-service/values.preview.template.yaml
+++ b/charts/rpe-send-letter-service/values.preview.template.yaml
@@ -1,0 +1,6 @@
+java:
+  # Don't modify below here
+  image: ${IMAGE_NAME}
+  ingressHost: ${SERVICE_FQDN}
+postgresql:
+  enabled: true
--- a/charts/rpe-send-letter-service/values.preview.template.yaml
+++ b/charts/rpe-send-letter-service/values.preview.template.yaml
@@ -2,5 +2,10 @@ java:
   # Don't modify below here
   image: ${IMAGE_NAME}
   ingressHost: ${SERVICE_FQDN}
+  environment:
+    FTP_USER: ${TEST_FTP_USER}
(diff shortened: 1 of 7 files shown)


charts/rpe-send-letter-service/values.preview.template.yaml
```diff
@@ -0,0 +1,6 @@
+java:
```
nitinprabhu: are we sure we don't need any env config here ?

charts/rpe-send-letter-service/values.yaml
```diff
@@ -0,0 +1,31 @@
...
+    FTP_PORT: "22"
+    FTP_FINGERPRINT: "SHA256:3tX3DIkqd1Loz2alHfnt+qjHocfxk0YUOZHlnf9Zgdk"
+    FTP_TARGET_FOLDER: "TO_XEROX"
+    FTP_REPORTS_FOLDER: "FROM_XEROX"
+    FTP_DOWNTIME_FROM: "23:58"
+    FTP_DOWNTIME_TO: "23:59"
+    # mails
+    SPRING_MAIL_HOST: "false"
+  # Don't modify below here
+  image: ${IMAGE_NAME}
```
doncem: shouldn't be hardcoded: https://github.com/hmcts/draft-store/blob/master/charts/draft-store-service/values.yaml#L12

charts/rpe-send-letter-service/values.yaml
```diff
@@ -0,0 +1,31 @@
...
+  ingressHost: ${SERVICE_FQDN}
```
doncem: shouldn't be set by default:
```suggestion
```

doncem:
will work a bit later on this

PR #560, opened 2019-09-27
Remove no longer needed config

rkondratowicz:
Removing config for functionality removed in https://github.com/hmcts/send-letter-service/pull/557


rkondratowicz:
@nitinprabhu It is getting removed in this PR.

PR #562, opened 2019-10-01
Don't read blob data when fetching pending letters from DB

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-836


src/main/java/uk/gov/hmcts/reform/sendletter/entity/BasicLetterInfo.java
```diff
@@ -10,6 +10,7 @@
     private String service;
     private String status;
     private String type;
+    private String encryptionKeyFingerprint;
```
rkondratowicz: reusing the same model as for 'stale letters'.
This field is expected by the pending letters model.

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LetterRepository.java
```diff
@@ -28,6 +28,13 @@
         @Param("createdBefore") LocalDateTime createdBefore
     );
 
+    @Query("select new uk.gov.hmcts.reform.sendletter.entity.BasicLetterInfo(l.id, l.checksum, l.service, l.status, l.type, l.encryptionKeyFingerprint l.createdAt, l.sentToPrintAt)"
+        + " from Letter l "
+        + " where l.status = 'Created'"
+        + " and l.type <> '" + UploadLettersTask.SMOKE_TEST_LETTER_TYPE + "'"
+        + " order by l.createdAt desc")
```
lgonczar: How about ascending order, so that the oldest letters would appear first?

lgonczar: Also, needs integration tests.

PR #564, opened 2019-10-01
Upload only one batch of files in single job run

rkondratowicz:
To prevent racing condition when lock expires and other instances start processing the same letters.


lgonczar:
Like discussed, better to let the job do its thing continuously and make it stop just before the lock timeout - the point we're typically not going to reach. In a typical scenario, the job doesn't take that long and doing one batch at a time would decrease throughput a bit.

PR #566, opened 2019-10-03
Read one letter from db at a time

rkondratowicz:
Also make it easier to change the batch size for job run

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
@@ -17,7 +17,6 @@ import uk.gov.hmcts.reform.sendletter.services.ftp.ServiceFolderMapping;
 import uk.gov.hmcts.reform.sendletter.services.util.FinalPackageFileNameHelper;
 
 import java.time.ZoneId;
-import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 
@@ -27,7 +26,9 @@ import static uk.gov.hmcts.reform.sendletter.util.TimeZones.EUROPE_LONDON;
 @Component
 @ConditionalOnProperty(value = "scheduling.enabled", matchIfMissing = true)
 public class UploadLettersTask {
+
     private static final Logger logger = LoggerFactory.getLogger(UploadLettersTask.class);
+    public static final int BATCH_SIZE = 10;
     public static final String SMOKE_TEST_LETTER_TYPE = "smoke_test";
     private static final String TASK_NAME = "UploadLetters";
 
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -51,73 +52,78 @@ public UploadLettersTask(
...
         logger.info("Started '{}' task", TASK_NAME);
 
-        List<Letter> lettersToUpload = repo.findFirst3ByStatus(LetterStatus.Created);
-        int counter = 0;
-
-        if (!lettersToUpload.isEmpty()) {
-            counter = ftp.runWith(sftpClient -> uploadLetters(lettersToUpload, sftpClient));
+        if (!availabilityChecker.isFtpAvailable(now(ZoneId.of(EUROPE_LONDON)).toLocalTime())) {
+            logger.info("Not processing '{}' task due to FTP downtime window", TASK_NAME);
+        } else {
```
doncem: now with this move makes it even more sensible to move `started this task` to aop component which we already have specific for schedules. just a blarb, not take care right now

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -51,73 +52,78 @@ public UploadLettersTask(
...
+                if (letter.isPresent()) {
+                    boolean uploaded = processLetter(letter.get(), client);
+                    if (uploaded) {
+                        uploadCount++;
+                    }
                 } else {
-                    letter.setStatus(LetterStatus.Skipped);
+                    break;
                 }
+            }
```
doncem: Bah, java9 here would be very handy with `.takeWhile` function

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
```diff
@@ -51,73 +52,78 @@ public UploadLettersTask(
...
+        return ftp.runWith(client -> {
+            int uploadCount = 0;
+
+            for (int i = 0; i < BATCH_SIZE; i++) {
+                Optional<Letter> letter = repo.findFirstByStatusOrderByCreatedAtAsc(LetterStatus.Created);
+                if (letter.isPresent()) {
+                    boolean uploaded = processLetter(letter.get(), client);
+                    if (uploaded) {
+                        uploadCount++;
+                    }
```
doncem: We could just use `if` statement alone without assigning new var

src/main/java/uk/gov/hmcts/reform/sendletter/tasks/UploadLettersTask.java
rkondratowicz: soon ;)

PR #573, opened 2019-10-16
Handle letter with number of document copies specified

rkondratowicz:
Only `PdfCreator` changes here.
`LetterService` and controller changes in next PRs.


src/main/java/uk/gov/hmcts/reform/sendletter/model/in/LetterWithPdfsAndCopiesRequest.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.List;
+import java.util.Map;
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.Size;
+
+public class LetterWithPdfsAndCopiesRequest implements ILetterRequest {
+
+    @Size(min = 1, max = 30)
+    public List<Doc> documents;
```
nitinprabhu: should this be final ?

rkondratowicz: yup. fixed.

lgonczar:
I think the most value in this story comes from the fact that we could avoid dealing with potentially large amounts of data in memory. For that to happen, we need to modify PdfMerger to accept a collection of streams and return a stream, and use it like this in PdfCreator.

lgonczar:
Discussion would have been nice

src/main/java/uk/gov/hmcts/reform/sendletter/services/pdf/PdfCreator.java
```diff
@@ -43,6 +46,19 @@ public PdfCreator(DuplexPreparator duplexPreparator, IHtmlToPdfConverter convert
...
     }
 
+    public byte[] createFromBase64PdfWithCopies(List<Doc> docs) {
+        Asserts.notNull(docs, "base64decodedDocs");
+
+        List<byte[]> pdfs = docs
+            .stream()
+            .map(d -> Collections.nCopies(d.copies, d.content))
+            .flatMap(Collection::stream)
+            .map(duplexPreparator::prepare)
```
mdayican: I think this we should do this before nCopies operations as this method creates new copies ???

PR #575, opened 2019-10-16
Update owasp suppressions

rkondratowicz:


config/owasp/suppressions.xml
```diff
@@ -25,19 +25,16 @@
     </suppress>
 
     <suppress>
```
nitinprabhu: let's put an end date  `<suppress  until="2019-10-31">` ?

PR #593, opened 2019-10-24
Update swagger spec

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/sendletter/controllers/ControllerResponseMessage.java
```diff
@@ -4,4 +4,5 @@
 final class ControllerResponseMessage {
 
     static final String RESPONSE_401 = "Invalid service authorisation header";
+    static final String RESPONSE_403 = "Service not configured";
```
mdayican: may be ,
"Service not authorised" 
"Service not configured to authorised" ?

rkondratowicz: Status itself gives the information that service is not allowed to perform this action.

lgonczar: Good point, `Service not authorised` might translate to our internal "not configured", but it would be ideal if the caller was told that they're just not authorised.

PR #596, opened 2019-10-25
disable ENCRYPTION

mdayican:

https://tools.hmcts.net/jira/browse/BPS-838

### Change description ###

disable ENCRYPTION for staging

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/rpe-send-letter-service/values.aat.template.yaml
+++ b/charts/rpe-send-letter-service/values.aat.template.yaml
@@ -3,3 +3,5 @@ java:
   image: ${IMAGE_NAME}
   ingressHost: ${SERVICE_FQDN}
   devApplicationInsightsInstrumentKey: "71df5a0f-ebbe-445c-b1d1-d10d349f0970"
+  environment:
+    ENCRYPTION_ENABLED: "false"


charts/rpe-send-letter-service/values.aat.template.yaml
```diff
@@ -3,3 +3,5 @@ java:
   image: ${IMAGE_NAME}
   ingressHost: ${SERVICE_FQDN}
   devApplicationInsightsInstrumentKey: "71df5a0f-ebbe-445c-b1d1-d10d349f0970"
+  environment:
+    ENCRYPTION_ENABLED: "false"
```
nitinprabhu: encryption is enabled on AAT ?

mdayican: for staging , in jenkins_cnp encryption is false so it should be false here as well.

nitinprabhu: This is used by the app ? In jenkins file we set for functional tests

mdayican: LetterService is using.

mdayican: in jenkins it is TEST_ENCRYPTION_ENABLED

mdayican:
> encryption should be ON



> encryption should be ON

it should be off as tests are expecting OFF.
in  jenkinsfile_CNP  env.TEST_ENCRYPTION_ENABLED = false

PR #597, opened 2019-10-25
enable encryption

mdayican:
### Change description ###

enable encryption for all env.

```
[ ] Yes
[x] No
```


mdayican:
> Could you please bump chart version ?

no need to bump as there is no change in values.yaml

PR #610, opened 2019-11-27
Simplify liveness alert check

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-933

Removing the `where` clause completely. Send letter consists of a single service currently.


infrastructure/liveness-alert.tf
```diff
@@ -10,7 +10,6 @@ module "send-letter-service-liveness-alert" {
   app_insights_query = <<EOF
 requests
 | where name == "GET /health" and resultCode != "200"
-| where url contains "send-letter-service"
```
rkondratowicz: no need to filter there are no other services logging to this app insights

PR #645, opened 2020-02-11
Fpla onboarding

mdayican:

https://tools.hmcts.net/jira/browse/BPS-1046

### Change description ###

Fpla config mapping

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -73,6 +73,8 @@ reports:
       display-name: SSCS
     - service: finrem_document_generator
       display-name: FINREM
+    - service: fpl_case_service
+      display-name: FPLA
 
 flyway:
   user: ${LETTER_TRACKING_DB_USER_NAME}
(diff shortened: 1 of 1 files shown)


infrastructure/.terraform-version
```diff
@@ -0,0 +1 @@
+0.12.20
```
nitinprabhu: change not related could you please create separate PR ?

infrastructure/main.tf
```diff
@@ -1,5 +1,5 @@
 provider "azurerm" {
-  version = "1.23.0"
+  version = "=1.42.0"
```
nitinprabhu: same here

src/main/resources/application.yaml
```diff
@@ -104,6 +106,8 @@ ftp:
       folder: SSCS
     - service: finrem_document_generator
       folder: FINREM
+    - service: fpl_case_service
+      folder: BULKPRINT
```
nitinprabhu: could add a comment here that this needs to be changed later ?

PR #648, opened 2020-02-13
Remove deprecated property in chart for preview template

doncem:
### Change description ###

TTL 14 days

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #649, opened 2020-02-13
Remove deprecated property in chart for preview template

doncem:
### Change description ###

Recreating #648 

The problem was with too old java chart version. Only from 2.13.* series this is enabled

```
[ ] Yes
[x] No
```


doncem:
should be covered by #654. Builds for this one also never worked :/

PR #671, opened 2020-03-24
Suppress/Revert exception throws for used to be silent rollbacks

doncem:
### Change description ###

Came along with #665 which has this feature: https://github.com/pgjdbc/pgjdbc/pull/1729

```
[ ] Yes
[x] No
```


doncem:
no longer needed

PR #672, opened 2020-03-24
Use JDBC lock provider instead of JDBC template 

nitinprabhu:
### Change description ###

- JDBC template with latest version of shed lock seems to have some issues with race condition of acquiring lock.

- This is an attempt to see if it fixes it. Processor already uses it and doesn't see any issues.

```
[ ] Yes
[ x] No
```


nitinprabhu:
Closing this for now but if we see the issue in future then we can reopen this one

PR #684, opened 2020-04-07
Add repo method for finding letters created on given day

rkondratowicz:
Also adding 'printed_at' to existing model.

https://tools.hmcts.net/jira/browse/BPS-1119

PR #686, opened 2020-04-08
Add endpoint for finding all letters from given day

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-1119


src/main/java/uk/gov/hmcts/reform/sendletter/controllers/reports/LetterListController.java
```diff
@@ -0,0 +1,45 @@
...
+public class LetterListController {
+
+    private final LetterRepository repository;
+
+    public LetterListController(LetterRepository repository) {
+        this.repository = repository;
+    }
+
+    @GetMapping(path = "/letters")
+    public LettersInfoResponse getLetters(@RequestParam(name = "date") @DateTimeFormat(iso = DATE) LocalDate date) {
```
nitinprabhu, line 28: is it worth adding optional `service` param so that requester can filter them ?

nitinprabhu: Also this would return bulk scan letter in prod used for smoke tests ?

rkondratowicz: smoke test letters are filtered out here: https://github.com/hmcts/send-letter-service/pull/684/files

rkondratowicz: >is it worth adding optional service param so that requester can filter them ?
It's not in the ticket, but good idea, let's see what business says after they see current solution

PR #729, opened 2020-06-02
Log bad requests

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/sendletter/ResponseExceptionHandler.java
```diff
@@ -57,7 +57,11 @@ protected void activateDirectFieldAccess(DataBinder dataBinder) {
                 .map(err -> new FieldError(err.getField(), err.getDefaultMessage()))
                 .collect(toList());
 
-        return badRequest().body(new ModelValidationError(fieldErrors));
+        ModelValidationError error = new ModelValidationError(fieldErrors);
+
+        log.info("Bad request: {}", error);
```
nitinprabhu, line 62: can this log base64 encoded documents ?

rkondratowicz: No, the only validation problem with documents (on this stage) can be that it's empty, or there are too many of them.

PR #757, opened 2020-07-16
Feature: postgres v11 database(s)

doncem:
### JIRA link (if applicable) ###

[Migrate send letter DB to version 11](https://tools.hmcts.net/jira/browse/BPS-1057)

### Change description ###

This PR:

- Creates new DB
- Stores it's secrets in vault
- Replaces staging db to v11. It is only used for testing purposes in pipeline. Can safely nuke it and rebuild from scratch

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #760, opened 2020-07-17
Add new db (v11) configs

doncem:

[Migrate send letter DB to version 11](https://tools.hmcts.net/jira/browse/BPS-1057)

### Change description ###

Follow-up #757 - bootstrap configs to app before starting to use them

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/bootstrap.yaml
+++ b/src/main/resources/bootstrap.yaml
@@ -9,6 +9,11 @@ spring:
         ftp-private-key: FTP_PRIVATE_KEY
         ftp-public-key: FTP_PUBLIC_KEY
         upload-summary-report-recipients: UPLOAD_SUMMARY_REPORT_RECIPIENTS
+        send-letter-service-db-database-v11: DB_NAME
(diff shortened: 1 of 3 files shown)


src/main/resources/bootstrap.yaml
```diff
@@ -9,6 +9,11 @@ spring:
         ftp-private-key: FTP_PRIVATE_KEY
         ftp-public-key: FTP_PUBLIC_KEY
         upload-summary-report-recipients: UPLOAD_SUMMARY_REPORT_RECIPIENTS
+        send-letter-service-db-database-v11: DB_NAME
```
nitinprabhu, line 12: so plan is later use this env vars in app yaml ?

doncem: doesn't really matter. but for compatibility i guess we can keep previous one. resources will be overridden after the complete migration. migration for now will support 2 data sources. once migration is complete - we'll just flip config and be off with new db

PR #761, opened 2020-07-17
Configure multiple (2nd) datasource for (spring) application

doncem:

[Migrate send letter DB to version 11](https://tools.hmcts.net/jira/browse/BPS-1057)

### Change description ###

Follow-up #760. It:

- adds extra set of configs for new db
- assures legacy db config (which is current one) is in use as a primary source

No new ones will be loaded as there are none - job for next PR

```
[ ] Yes
[x] No
```


doncem:
doesn't want to play ball no matter what i tried. time pressure draws the line and we will have to go with initial solution - downtime

PR #765, opened 2020-07-20
Add scheduled job for clearing content of old letters

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-1328

Flux changes: https://github.com/hmcts/cnp-flux-config/pull/4832

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -126,6 +126,11 @@ file-cleanup:
   cron: ${FILE_CLEANUP_CRON:0 15 * * * *}
   ttl: 12h
 
+letter-content-cleanup:
+  enabled: ${LETTER_CONTENT_CLEANUP_ENABLED:false}
+  cron: ${LETTER_CONTENT_CLEANUP_CRON:0 0 7 * * MON}
+  ttl: P31D
+
 encryption:
   enabled: ${ENCRYPTION_ENABLED:false}
   publicKey: ${ENCRYPTION_PUBLIC_KEY:}
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -126,9 +126,9 @@ file-cleanup:
   cron: ${FILE_CLEANUP_CRON:0 15 * * * *}
   ttl: 12h
 
-letter-content-cleanup:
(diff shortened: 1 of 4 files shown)


src/main/resources/application.yaml
```diff
@@ -126,6 +126,11 @@ file-cleanup:
   cron: ${FILE_CLEANUP_CRON:0 15 * * * *}
   ttl: 12h
 
+letter-content-cleanup:
```
lgonczar, line 129: Maybe it would be worth mentioning somewhere in the name that it's about unposted letters, because we'll still cleanup letter content, just after it's confirm as posted.

src/test/java/uk/gov/hmcts/reform/sendletter/tasks/ClearOldLetterContentTaskTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Mock LetterRepository letterRepo;
+    @Mock Clock clock;
+
+    @Test
+    void should_call_repo_to_clear_uploaded_letters() {
+        // given
+        var now = LocalDateTime.now();
+        given(clock.instant()).willReturn(now.toInstant(UTC));
+        given(clock.getZone()).willReturn(UTC);
```
lgonczar, line 29: This UTC is a bit unclear, since we're operating on Europe/London

src/main/resources/application.yaml
```diff
@@ -126,6 +126,11 @@ file-cleanup:
   cron: ${FILE_CLEANUP_CRON:0 15 * * * *}
   ttl: 12h
 
+old-letter-content-cleanup:
+  enabled: ${OLD_LETTER_CONTENT_CLEANUP_ENABLED:false}
```
nitinprabhu, line 130: this job won't be running in AAT staging which is fine I think as you don't have any func tests relying on the job.

Otherwise `OLD_LETTER_CONTENT_CLEANUP_ENABLED` should be defined in values yaml with false value and then overrided wherever needed.

rkondratowicz: Yup, I've made this config match the other AAT job (`DeleteOldFilesTask.java` that uses `FILE_CLEANUP_ENABLED` env var)

PR #954, opened 2021-01-07
Add no fault divorce API to config

linusnorton:
The no fault divorce case orchestration service is a fork of the divorce one. It has a new microservice name so we need to add the mapping. For now it uses the same config/structure as divorce but we may need to change it in the future


jasonpaige (approved):
🚀

PR #1204, opened 2021-07-26
DTSBPS-686: adding fortify scan

rishikrsharma:
DTSBPS-686

### Change description ###
adding fortify scan

```
[ ] Yes
[x ] No
```

### DIFF
--- a/config/fortify-client.properties
+++ b/config/fortify-client.properties
@@ -1,0 +1,1 @@
+fortify.client.releaseId=65430
--- a/config/fortify-client.properties
+++ b/config/fortify-client.properties
@@ -1,1 +1,1 @@
-fortify.client.releaseId=65430
+fortify.client.releaseId=65508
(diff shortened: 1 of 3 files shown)


jasonpaige (approved):

config/fortify-client.properties
```diff
@@ -0,0 +1 @@
+fortify.client.releaseId=65430
```
rajaram-nagarajan, line 1: You need to use the release id of send-letter-service here. ID should be 65508, but better to double check.

rajaram-nagarajan (changes requested):

rajaram-nagarajan (changes requested):
Also, change your branch name to start with 'nightly' (ex: nightly-fod-scan-dtsbps-686) to easily understand the task by its name in jenkins.

config/fortify-client.properties
rishikrsharma: I have updated the SLS release id

rishikrsharma:
> nightly-fod-scan-dtsbps-686

sure

PR #1398, opened 2022-02-02
grants access to storage account from perftest mgmt subnet

jasonpaige:

```
[ ] Yes
[X] No
```


rishikrsharma (approved):

infrastructure/perftest_mgmt_subnet.tf
```diff
@@ -0,0 +1,12 @@
+locals {
+  perftest_vnet_name           = "core-infra-vnet-perftest"
```
mdayican, line 2: why do you need this?

jasonpaige: It's required by `azurerm_subnet` (see below).
The perftest mgmt subnet needs access to the storage account to either prime it with data or capture results at the end of a perfetst run.

PR #1605, opened 2022-07-01
Return 'Bad request' when letter is posted (DTSBPS-995)

alivenichoppa:
https://tools.hmcts.net/jira/browse/DTSBPS-995

### Change description ###
Update mark-aborted endpoint to return Bad Request for 'Posted' letters.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/sendletter/exception/UnsupportedLetterStatusException.java
```diff
@@ -0,0 +1,16 @@
+package uk.gov.hmcts.reform.sendletter.exception;
+
+import uk.gov.hmcts.reform.sendletter.entity.LetterStatus;
+
+import java.util.UUID;
+
+public class UnsupportedLetterStatusException extends RuntimeException {
```
apereverzin, line 7: I suggest to rename this to WrongLetterStatusException (and reflect this in the message in line 10) because semantically the status is supported but is wrong in this particular situation.

alivenichoppa: I want to say that the letter status is not supported for aborting the letter. I renamed it as `UnableToAbortLetterException`

apereverzin (approved):

PR #1787, opened 2022-09-22
Add config for civil-service (DTSBPS-1079)

alivenichoppa:
https://tools.hmcts.net/jira/browse/DTSBPS-1079

### Change description ###
Whitelist civil-service

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -114,6 +114,7 @@ ftp:
       folder: CMC
     - service: civil-service
       folder: CMC
+      enabled: ${CIVIL_SERVICE_ENABLED:false}
     - service: send_letter_tests
       folder: BULKPRINT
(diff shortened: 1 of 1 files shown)


src/main/resources/application.yaml
```diff
@@ -71,6 +71,8 @@ reports:
   service-config:
     - service: cmc_claim_store
       display-name: CMC
+    - service: civil-service
+      display-name: CMC
```
apereverzin, line 75: Are you sure display-name should be the same?

apereverzin (approved):

src/main/resources/application.yaml
alivenichoppa: raised https://github.com/hmcts/send-letter-service/pull/1788

PR #2154, opened 2023-03-07
DTSBPS-1322: repository layer

justiceia:
https://tools.hmcts.net/jira/browse/DTSBPS-1322

### Change description ###
repository layer changes

```
[ ] Yes
[ x ] No
```


src/main/java/uk/gov/hmcts/reform/sendletter/entity/DocumentRepository.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.Optional;
+import java.util.UUID;
+
+@SuppressWarnings("checkstyle:LineLength")
+public interface DocumentRepository extends JpaRepository<Document, UUID> {
+
+    @Query(value = "select * from documents d "
+        + " WHERE d.checksum = :checkSum "
+        + " AND d.created_at >= :createdAfter limit 1", nativeQuery = true)
+    Optional<Document> findCreatedAfter(@Param("checkSum") String checkSum, @Param("createdAfter") LocalDateTime createdAfter);
```
amstevenson, line 17: Looks good, a quick question though

For the findCreatedAfter part of the JPA repo, was this where we were going to get a list of documents back between the current time minus a configurable time (one hour for example)?

I see a test for the current time minus 30 mins, so that'll be the configurable part I imagine, so thats all good :slightly_smiling_face: 

Just wondering if we want to bring back a list of documents instead? 

If we get a request with more than one document and we want to check if one or more have checksums on that table, we could check all of them in one go with an IN statement on the sql query

I.e get all documents where the checksum is in the ones passed through in the request (one or more)

apereverzin: We just need to find if any document with the same checkSum and creted after the cutoff timestamp exists. This method can be renamed to 'findOneCreatedAfter'. Ultimately it should return boolean value - exists or not, returning the optional result is a kind of a compromise.

And there is no requirement to retrieve the list of the documents with the same checksum.

apereverzin (approved):

PR #2512, opened 2024-04-04
Fact 1464

amstevenson:
Removal of a 404 error each time a letter is saved. Tries up to three times as opposed to always displaying one in log analytics. 

https://tools.hmcts.net/jira/browse/FACT-1464


RuthKirby (approved):
👍

PR #2523, opened 2024-04-22
Adds import.tf file

RuthKirby:
### Change description ###

Adds import.tf file to try and fix demo pipeline by importing demo secrets into state

```
[ ] Yes
[x] No
```


infrastructure/import.tf
```diff
@@ -0,0 +1,30 @@
...
+    },
+    {
+      id = "https://rpe-send-letter-demo.vault.azure.net/secrets/send-letter-service-POSTGRES-DATABASE/ee565b76f0614b81936ca93c8ebc9285"
+      to = azurerm_key_vault_secret.flexible_secret_standard_format["DATABASE"]
+    }
+  ]
+}
+
+import {
+  for_each = { for import in local.demo_imports : import if var.env == "demo" }
```
MartyFox, line 27: The error is: `Key expression is required when building an object.` which applies to this line.
In a `for_each` you can use `[]` for lists or `{}` for objects and using `{}` requires keys to be set e.g

```
import {
  for_each = { for import in local.demo_imports : import.id => import if var.env == "demo" }
  id = each.key
  to = each.value.to
}

PR #2616, opened 2024-09-23
FACT-1660 Removed unused alerts

joshblackmoor:

https://tools.hmcts.net/jira/browse/FACT-1660

### Change description ###

Remove unused alerts, covered by different monitoring tools 

```
[ ] Yes
[x] No
```


RuthKirby (commented):
You may be able to get rid of this ([FtpFileUploaded](https://github.com/hmcts/send-letter-service/blob/b32cc07116871211b637c12c2c8fac16f70298a5/src/main/java/uk/gov/hmcts/reform/sendletter/services/ftp/FtpClient.java#L72)) dependency log from the application code then as this is_down.tf alert is what it is being used for. It is the last dependency that spams a lot in Send Letter Service.

RuthKirby (approved):

joshblackmoor:
> dependency

Good point! Have raised https://tools.hmcts.net/jira/browse/FACT-2010 for us to remove them as there are quite a few of those types of logs and would need proper time to check which ones we should/shouldn't remove

PR #2761, opened 2026-02-12
FACT-2678 - abort letter status changes

hurricanepilot:

[FACT-2678](https://tools.hmcts.net/jira/browse/FACT-2678)

### Change description ###

- Moved the report processing task into a service that will be triggered by the bsp cron trigger project.
- Adding support for moving a letter into the NO_REPORT_ABORTED status as part of a service that is triggered by the bsp cron trigger project

**Does this PR introduce a breaking change?**

```
[ ] Yes
[x] No
```


src/main/resources/application.yaml
josh-uk: Can we remove this now we've removed the task or is it being used elsewhere? 

(I think if this can go we may likely need to change the chart if its being set in there too

josh-uk (approved):

josh-uk (approved):

PR #2767, opened 2026-04-16
Add API check for reports

RebeccaHayleyPickles:
FACT-2679

Add missing report check

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/sendletter/controllers/reports/ReportsController.java
```diff
@@ -71,4 +73,27 @@ public ResponseEntity<byte[]> getCountSummary(
...
+     */
+    @GetMapping(path = "/check-reports")
+    @Operation(description = "Checks if all reports for a date range are present")
+    public ResponseEntity<List<MissingReportsResponse>> checkReports(
+        @RequestParam(name = "startDate") @DateTimeFormat(iso = DATE) LocalDate startDate,
+        @RequestParam(name = "endDate") @DateTimeFormat(iso = DATE) LocalDate endDate
+    ) {
+        List<MissingReportsResponse> missingReports = reportsService.checkReports(startDate, endDate);
+
+        if (missingReports.isEmpty()) {
```
amstevenson: you can use a ternary expression here to make it one line if you want to. Would look cleaner, but up to you

src/main/java/uk/gov/hmcts/reform/sendletter/controllers/reports/ReportsController.java
```diff
@@ -71,4 +73,27 @@ public ResponseEntity<byte[]> getCountSummary(
...
+     *
+     * @param startDate The start date of the range to check
+     * @param endDate   The end date of the range to check
+     * @return ResponseEntity with HTTP 200 if all reports are present, or HTTP 404 with list of missing reports
+     */
+    @GetMapping(path = "/check-reports")
+    @Operation(description = "Checks if all reports for a date range are present")
+    public ResponseEntity<List<MissingReportsResponse>> checkReports(
+        @RequestParam(name = "startDate") @DateTimeFormat(iso = DATE) LocalDate startDate,
+        @RequestParam(name = "endDate") @DateTimeFormat(iso = DATE) LocalDate endDate
```
amstevenson: I presume start and end date from the cron will be the seven days mentioned in the ticket. 

We could default if the end date is missing for it to be today, potentially. Or also default the first to also be seven days previous to today. 

Wouldn't worry about it though. Not so much a point to address but food for thought for you :)

src/main/java/uk/gov/hmcts/reform/sendletter/services/ReportsService.java
```diff
@@ -76,6 +86,33 @@ public List<LettersCountSummary> getCountFor(LocalDate date) {
...
+    public List<MissingReportsResponse> checkReports(
+        LocalDate startDate,
+        LocalDate endDate
+    ) {
+        Set<String> expectedCodes = reportsServiceConfig.getReportCodes();
+        List<Report> actualReports = reportRepository.findByReportDateBetween(startDate, endDate);
+
+        List<MissingReportsResponse> missing = new ArrayList<>();
+
+        for (LocalDate date = startDate; !date.isAfter(endDate); date = date.plusDays(1)) {
```
amstevenson: I think this will go through the reports a fair few times, unless I am mistaken. 

for days -> for codes -> for each reports
pretty much a days x codes then two reports?

Correct me if I am wrong :D I probably am

RebeccaHayleyPickles: Yes it does, but unfortunately each scenario needs to be checked. I've tried to improve on it by only pulling the relevant reports from the db and changing the checks to only pull the relevant info

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/controllers/reports/CheckReportsTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @Autowired
+    private ReportsServiceConfig reportsServiceConfig;
+
+    @MockitoBean
+    private AuthTokenValidator tokenValidator;
+
+    @AfterEach
+    void tearDown() {
+        reportRepository.deleteAll();
```
amstevenson, line 47: I think we use test containers for integration tests? If so, can likely remove this and use different data per test. Then it'll auto teardown afterwards. 

Do correct me if I am wrong though.

Edit: I see that perhaps its to do with the database state after each, before the endpoint is called. In which case it may be merited

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/controllers/reports/CheckReportsTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @Test
+    void should_return_404_when_reports_are_missing() throws Exception {
+        // given
+        LocalDate startDate = LocalDate.of(2026, 1, 1);
+        LocalDate endDate = LocalDate.of(2026, 1, 1);
+        Set<String> services = reportsServiceConfig.getServiceConfig().keySet();
+        String service = services.iterator().next(); // Use the first service
+
+        letterRepository.save(new Letter(
```
amstevenson, line 123: Perhaps add a success one in here too? As it should still give us an error at the end. Just in case there is some weird edge case

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LettersCountSummaryRepository.java
```diff
@@ -32,4 +33,33 @@ public interface LettersCountSummaryRepository extends JpaRepository<Letter, UUI
...
+     *
+     * @param dateFrom the date from
+     * @param dateTo   the date to
+     * @return the list
+     */
+    @Query(
+        nativeQuery = true,
+        value = "SELECT DISTINCT \n"
+            + "  created_at::date as \"createdAt\", \n"
+            + "  (CASE \n"
```
amstevenson, line 48: I think here, you have a case within a case, you could probably use a flattened case

src/main/java/uk/gov/hmcts/reform/sendletter/entity/LettersCountSummaryRepository.java
```diff
@@ -32,4 +33,33 @@ public interface LettersCountSummaryRepository extends JpaRepository<Letter, UUI
...
+            + "    WHEN service = 'sscs' THEN (CASE \n"
+            + "      WHEN additional_data->>'isIbca' = 'true' THEN 'sscs-ib' \n"
+            + "        ELSE 'sscs-reform' END) \n"
+            + "      ELSE service \n"
+            + "  END) as service,\n"
+            + "  COALESCE(CAST(additional_data->>'isInternational' AS boolean), false) as international\n"
+            + " FROM letters \n"
+            + " WHERE status IN "
+            + "('Created', 'Uploaded', 'FailedToUpload', 'Posted', 'PostedLocally', 'Aborted', 'NotSent')\n"
+            + "  AND type <> 'smoke_test'\n"
```
amstevenson: Can do type IS DISTINCT FROM 'smoke_test'. Which handles nulls...I believe. I am not sure if type can be null...so

src/main/java/uk/gov/hmcts/reform/sendletter/services/ReportsService.java
```diff
@@ -76,6 +86,34 @@ public List<LettersCountSummary> getCountFor(LocalDate date) {
             .collect(toList());
     }
 
+    public List<MissingReportsResponse> checkReports(
+        LocalDate startDate,
+        LocalDate endDate
+    ) {
+        List<ServiceLettersReport> serviceLetters = repo.getServiceLettersReport(
+            startDate.atStartOfDay(),
+            endDate.atTime(LocalTime.MAX)
```
amstevenson: I am not sure what LocalTime.MAX is. Is this going to be taking minutes/seconds into account?

May be safer to do something like endDate.plusDays(1).atStartOfDay(), and then change the sql to be created_at < :dateTo rather than created_at <= :dateTo

Although I am just thinking out loud here, best to double check!

src/main/java/uk/gov/hmcts/reform/sendletter/services/ReportsService.java
```diff
@@ -76,6 +86,34 @@ public List<LettersCountSummary> getCountFor(LocalDate date) {
...
+            .stream()
+            .map(r -> r.getReportDate() + r.getReportCode() + r.isInternational())
+            .collect(Collectors.toSet());
+
+        return serviceLetters.stream()
+            .map(sl -> {
+                String service = sl.getService();
+                String reportCode = reportsServiceConfig.getReportCode(service, null);
+                if (reportCode == null && service != null && service.startsWith("sscs-")) {
+                    reportCode = "SSCS-" + service.substring(5).toUpperCase();
```
amstevenson: Can you explain the substring 5 bit? Not sure if this pertains to the report at all. I guess its an edge case for SSCS? Interesting...just wondering why its needed

src/main/java/uk/gov/hmcts/reform/sendletter/services/ReportsService.java
RebeccaHayleyPickles: LocalTime.MAX is the maximum local time (taking into account the smallest time unit being used by LocalTime:  23:59:59.999999999. https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/time/LocalTime.html#MAX 

I can change it if you think it would be better to use .atStartOfDay()?

src/integrationTest/java/uk/gov/hmcts/reform/sendletter/controllers/reports/CheckReportsTest.java
RebeccaHayleyPickles: Yeah, the test container renews for each test session, but it doesn't clear the data after each test

src/main/java/uk/gov/hmcts/reform/sendletter/services/ReportsService.java
RebeccaHayleyPickles: Yep, it gets the rest of the substring of the sscs service name. It's because the report config 'getReportCode' needs both the service name and the letter status to work out what the rest of the report code is after 'sscs-' , but I can't get the letter status from the sql query because it would require pulling the letter object as well, which would slow it all down. If I take it out and expect the config to handle it, it just returns a null report code and the report is never checked. It's a work around because I think (@hurricanepilot understands it better) the formatting isn't consistent on the naming

src/main/java/uk/gov/hmcts/reform/sendletter/services/ReportsService.java
RebeccaHayleyPickles: I've changed it. I think it reads better

amstevenson (approved):

PR #2769, opened 2026-04-29
FACT-2750

amstevenson:
https://tools.hmcts.net/jira/browse/FACT-2750

### Change description ###
Checks: 
- Checked that the application can spin up with the new config. Even without the folder being present on the SFTP server
- Checked that the letter can be created based on the config and s2s setup/application configuration
- Checked that the letter cannot be created if the s2s config is not correct. I.e. incorrect key value
- Checked that if the folder does not exist, that it fails on upload
- Checked that the file does upload if the PCS folder exists
- Retention periods added as discussed

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -239,4 +239,4 @@ delete-old-letters:
   probate-backend-interval: ${PROBATE_BACKEND_INTERVAL:1 year}
   send-letter-tests-interval: ${SEND_LETTER_TESTS_INTERVAL:2 years}
   sscs-interval: ${SSCS_INTERVAL:3 months}
-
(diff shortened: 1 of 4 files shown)


src/main/resources/application.yaml
```diff
@@ -235,4 +240,3 @@ delete-old-letters:
   send-letter-tests-interval: ${SEND_LETTER_TESTS_INTERVAL:2 years}
   sscs-interval: ${SSCS_INTERVAL:3 months}
```
josh-uk, line 241: We need to know retention for the new service

hurricanepilot (approved):

benouaer (approved):

josh-uk (approved):

